PR #32, opened 2024-12-20
Add initial changelog

craigwalton-dsit:
Structure copied from inspect_ai.

To codify previous UK AISI, discussions we don't plan on publishing this package to PyPI until we've got some more mileage and consider it "stable".

I'm undecided on whether we should still be doing discrete "releases" (i.e. bumping package version, adding a tag, moving "unreleased" changelog items). Sometimes it is hard for us or users to reason about what version of the package they were using. Happy for opinions/input from others!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,7 @@
+# Changelog
+
+## Unreleased
+
+## 0.1.0 (19 December 2024)
+
+- Initial release.


CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog
+
+## Unreleased
```
craigwalton-dsit, line 3: I've intentionally not added a 0.1.0 section

```md
## 0.1.0 (19 December 2024)

- Initial release
```

because:
* we haven't really released (in the PyPI sense) a proper package
* I worry that users will see they're on 0.1.0 and assume they don't have any of the updates from "Unreleased"

jasongwartz-aisi (approved):
Happy to get this in so we've got something to work from!

I think whether or not to update this is sort of separate to whether or not we make a PyPI release - it's more about whether or not we're doing semver bumps/tags or not. 

With the docs currently telling people to add a git dependency, there's not really a version number specified, and the behaviour is up to the package manager; Poetry, for example, will install and lock the commit hash at time of install. So as long as we are instructing people to run eg. `poetry add git+https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox.git`, they'll be installing latest `HEAD`.

Fine to get this merged, but once we start asking people to install a semver release (either via eg. git tags, or the PyPI release) that would probably be the better time to start keeping this up-to-date.

craigwalton-dsit:
Good points! My thinking was to start updating this immediately, even though everything will just be under the "unreleased" header.

It won't be helpful for determining "does the version (i.e. commit) I have installed have x functionality?". But it will be helpful for:
- Getting into the habit of recording noteworthy changes
- Having a curated list of changes which users can refer to (rather than needing to look through PRs)

Issue #7, opened 2024-12-19
Support `inspect sandbox cleanup k8s` (with no params) to uninstall all releases


craigwalton-dsit:
This command ought to uninstall all Helm releases. Historically, we were hesitant to implement this as users may have been in a shared namespace.

Now, we encourage you to set up your cluster with user-specific namespaces, so this is less of a concern.

Might still be worth seeing if there's an easy way in Inspect to add a confirmation prompt/dialog when the user runs this though. The message might tell them "Warning: you're about to uninstall all (123) Helm charts in the current namespace."

craigwalton-dsit:
There is an easy way! https://inspect.ai-safety-institute.org.uk/interactivity.html#prompts

> I would recommend using the click prompt stuff, described a bit here: https://inspect.ai-safety-institute.org.uk/interactivity.html#prompts
Note that you don't need to use it as shown there (with input_screen) you should just use click directly
(input_screen is used when you are running inside an eval)

PR #45, opened 2025-01-03
Add support for `inspect sandbox cleanup k8s`

craigwalton-dsit:
Users often ask how to do this, and we normally have to share a command like `helm list -q | xargs -P 10 helm uninstall`.

Note that the [docs recommend](https://k8s-sandbox.ai-safety-institute.org.uk/getting-started/remote-cluster/#recommendations) setting up your cluster such that each user has their own namespace to reduce the risk of a careless user uninstalling someone else's Helm charts.

https://github.com/user-attachments/assets/03a02564-f4a2-40ff-a823-a01f97abdc5d

See issue #7.
See [Inspect's docs](https://inspect.ai-safety-institute.org.uk/sandboxing.html#environment-cleanup) on this command.

### DIFF
--- a/src/k8s_sandbox/_helm.py
+++ b/src/k8s_sandbox/_helm.py
@@ -70,7 +70,7 @@ class Release:
                     await asyncio.sleep(INSTALL_RETRY_DELAY_SECONDS)
 
     async def uninstall(self, quiet: bool) -> None:
-        await uninstall(self.release_name, quiet)
+        await uninstall(self.release_name, self._namespace, quiet)
 
     async def get_sandbox_pods(self) -> dict[str, Pod]:
         client = k8s_client()
@@ -121,6 +121,9 @@ class Release:
                 # Annotation do not have strict length reqs. Quoting/escaping
                 # handled by asyncio.create_subprocess_exec.
                 f"annotations.inspectTaskName={self.task_name}",
+                # Include a label to identify releases created by Inspect.
+                "--labels",
+                "inspectSandbox=true",
             ]
             + values,
             capture_output=True,
@@ -148,8 +151,7 @@ class Release:
         )
 
 
-async def uninstall(release_name: str, quiet: bool) -> None:
-    namespace = get_current_context_namespace()
+async def uninstall(release_name: str, namespace: str, quiet: bool) -> None:
     async with _uninstall_semaphore():
         sandbox_log(
             "Uninstalling helm release.", release=release_name, namespace=namespace
@@ -172,6 +174,24 @@ async def uninstall(release_name: str, quiet: bool) -> None:
         _raise_runtime_error(
             "Helm uninstall failed.", release=release_name, result=captured_output
         )
+
+
+async def get_all_release_names(namespace: str) -> list[str]:
+    result = await _run_subprocess(
+        "helm",
+        [
+            "list",
+            "--namespace",
+            namespace,
+            "-q",
+            "--selector",
+            "inspectSandbox=true",
+            "--max",
+            "0",
+        ],
+        capture_output=True,
+    )
+    return result.stdout.splitlines()
 
 
 def _raise_runtime_error(
--- a/src/k8s_sandbox/_sandbox_environment.py
+++ b/src/k8s_sandbox/_sandbox_environment.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import tempfile
 from contextlib import contextmanager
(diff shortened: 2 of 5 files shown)


src/k8s_sandbox/_helm.py
```diff
@@ -174,6 +176,24 @@ async def uninstall(release_name: str, quiet: bool) -> None:
...
+    result = await _run_subprocess(
+        "helm",
+        [
+            "list",
+            "--namespace",
+            namespace,
+            "-q",
+            "--selector",
+            "inspectSandbox=true",
+            "--max",
```
craigwalton-dsit, line 189: The max by default is 256 which has sometimes caught me out.

test/k8s_sandbox/inspect_integration/test_cleanup.py
```diff
@@ -40,4 +56,25 @@ def test_without_cleanup() -> None:
...
+    ):
+        run_and_verify_inspect_eval(task=task, model=model, sandbox_cleanup=False)
+
+    with patch("k8s_sandbox._helm.uninstall", wraps=uninstall) as spy:
+        # We don't want to actually uninstall all releases in this test (the test could
+        # be run on a production cluster).
+        with patch("rich.prompt.Confirm.ask", return_value=False) as confirm:
+            asyncio.run(K8sSandboxEnvironment.cli_cleanup(id=None))
+
+    assert "Are you sure you want to uninstall ALL" in confirm.call_args.args[0]
```
craigwalton-dsit, line 78: The fact that `confirm` has been called implies that at least 1 helm release was found.

src/k8s_sandbox/_helm.py
```diff
@@ -121,6 +121,9 @@ async def _install(self, upgrade: bool) -> None:
                 # Annotation do not have strict length reqs. Quoting/escaping
                 # handled by asyncio.create_subprocess_exec.
                 f"annotations.inspectTaskName={self.task_name}",
+                # Include a label to identify releases created by Inspect.
+                "--labels",
+                "inspectSandbox=true",
```
craigwalton-dsit, line 126: Hopefully this is release-level label key is specific enough. For other label keys in this project, we've used `inspect/service` (on Pods).

test/k8s_sandbox/inspect_integration/test_cleanup.py
```diff
@@ -40,4 +56,25 @@ def test_without_cleanup() -> None:
...
+    release = "no-clean"
+    with patch(
+        "k8s_sandbox._helm.Release._generate_release_name",
+        return_value=release,
+    ):
+        run_and_verify_inspect_eval(task=task, model=model, sandbox_cleanup=False)
+
+    with patch("k8s_sandbox._helm.uninstall", wraps=uninstall) as spy:
+        # We don't want to actually uninstall all releases in this test (the test could
+        # be run on a production cluster).
```
craigwalton-dsit, line 74: It might be nice if we added a requirement that the current namespace is a test-specific namespace where we're free to uninstall all releases. But weighed against the convenience of being able to run the test suite locally against the production cluster without creating/switching namespaces, I don't think it is currently worth it.

src/k8s_sandbox/_sandbox_environment.py
```diff
@@ -257,3 +257,21 @@ def validate_chart_dir(chart: Path | None) -> None:
...
+
+
+async def _uninstall_all_unmanaged_releases():
+    namespace = get_current_context_namespace()
+    releases = await get_all_release_names(namespace)
+    if len(releases) == 0:
+        print(f"No Inspect sandbox releases found in '{namespace}' namespace.")
+        return
+    if not Confirm.ask(
+        f"Are you sure you want to uninstall ALL {len(releases)} Inspect sandbox "
```
tcatling, line 269: I think it might be helpful to print the names out here. We could truncate this if there are a lot of releases, but i think even in that case it might be better to explicitly tell the user which things are going to be removed, as this is a pretty destructive action

craigwalton-dsit: This is a good point. In most cases, this will just be a list of random 8-character strings, but I think giving the user all the info they need to make an informed decision is a good shout.
Re truncation, yes I can see that having potentially hundreds of lines printed to the terminal could be annoying, but the user is about to do something potentially dangerous, so some terminal spam is a cheap price to pay and we certainly don't want any surprises or to mislead the user about what we're about to do.

This is why I capture the list of releases before showing the prompt - I don't want them to be surprised by a race condition between when we show the count in the prompt and when we actually action uninstalling things.

craigwalton-dsit: I've updated it to display a table of the releases to be uninstalled. I've chosen to display this before the prompt so that the prompt is still visible at the end should the list be very long.
Screenshot 2025-01-06 at 15 02 57

tcatling: nice!

tcatling (approved):
Looks great

Issue #29, opened 2024-12-19
Add basic info to docs site about how to configure NPs/CNPs to avoid performance/scaling issues


craigwalton-dsit:
Based on our experience with onboarding certain evals. @tcatling and @alex-remedios-aisi have a very good understanding of this.

PR #46, opened 2025-01-06
Add networking diagnostic test

craigwalton-dsit:
As suggested by Jason, let's put this script somewhere such that we can easily refer back to it and share with others. I haven't included the results as these are quite specific to our infrastructure at the time of running the tests and were not done in a particularly well controlled environment.

It will be run at a very small scale as part of CI just to prevent regressions in the script. I don't believe we'll get meaningful data from trying to run this at scale in CI:
1. We're heavily limited on how many concurrent deployments we can have on the default GitHub runner
2. There are no cluster-wide Cilium Network Policies (CNP) on our Minikube cluster (which we think were a contributor to the issues)

I've created an issue to better document our findings #29 and recommendations for CNPs.

### DIFF
--- a/test/diagnostics/network-issues/run.py
+++ b/test/diagnostics/network-issues/run.py
@@ -21,24 +21,24 @@ domains = ["google.com", "yahoo.com", "bing.com", "wikipedia.org", "amazon.com"]
 
 
 @task
-def internet_access_task():
+def internet_access_task(post_curl_sleep: int):
     return Task(
         dataset=MemoryDataset([Sample(input="Input", target=success_str)]),
         sandbox=("k8s", "helm-values.yaml"),
-        solver=[internet_access_solver()],
+        solver=[internet_access_solver(post_curl_sleep)],
         scorer=includes(),
     )
 
 
 @solver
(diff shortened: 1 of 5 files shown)


test/diagnostics/network-issues/run.py
```diff
@@ -0,0 +1,79 @@
...
+    os.environ["INSPECT_MAX_HELM_INSTALL"] = str(max_helm_install)
+    os.environ["INSPECT_MAX_HELM_UNINSTALL"] = str(max_helm_uninstall)
+    os.environ["INSPECT_MAX_POD_OPS"] = str(max_pod_ops)
+    logs = eval(
+        tasks=[internet_access_task(post_curl_sleep)],
+        model="mockllm/model",
+        max_samples=epochs,  # Let all epochs run concurrently.
+        epochs=epochs,
+    )
+    assert logs[0].results is not None
```
craigwalton-dsit, line 74: Just to satisfy Mypy.

jasongwartz-aisi (approved):
Thanks for adding this, and writing the doc 💯

PR #47, opened 2025-01-07
Improve Cilium Network Policy descriptions

craigwalton-dsit:
Housekeeping: whilst considering an update to Cilium on our cluster, I realised the CNP descriptions were a bit inconsistent and out of date wrt supporting `allowCIDR` and `allowEntities`.

### DIFF
--- a/src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
+++ b/src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
@@ -8,7 +8,7 @@ spec:
   description: |
     Allow egress to:
     - cluster-wide DNS (kube-dns),
-    - everything in the sandbox,
+    - all Pods in the same agent sandbox,
     - any configured allowDomains, allowEntities, or allowCIDR.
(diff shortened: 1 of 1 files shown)


src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
```diff
@@ -6,7 +6,10 @@ metadata:
     {{- toYaml $.Values.annotations | nindent 4 }}
 spec:
   description: |
-    Allow egress only to cluster-wide DNS, everything in the sandbox, and any allowDomains.
+    Allow egress to:
+    - cluster-wide DNS (kube-dns),
+    - all Pods in the same agent sandbox,
```
craigwalton-dsit, line 11: I realise that "endpoints" might be more accurate than "Pods", but I think Pods is more useful to a reader and is now consistent with the other CNP descriptions below.

alex-treebeard (approved):
all makes sense!

alex-remedios-aisi (approved):

PR #55, opened 2025-01-15
Add `unset` magic string for runtimeClassName

craigwalton-dsit:
We've been heavily opinionated that gVisor should be the container runtime of choice, but users can always override it if needed. However, you must override it to some non-null value. We'd like to add an escape hatch for users who wish to use their cluster's default runtime. We generally don't recommend taking this approach as this makes evals portable (now documented on docs site).

See Slack discussion https://inspectcommunity.slack.com/archives/C0855KJ6BMW/p1736847936476909

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Add `unset` magic string for `runtimeClassName` which will remove the field from the pod spec.
+- Add `CLUSTER_DEFAULT` magic string for `runtimeClassName` which will remove the field from the pod spec.
 - Add ignored `timeout_retry` parameter to `exec()` method.
 - Always capture the output of `helm uninstall` so that errors can contain meaningful information.
 - Add support for `inspect sandbox cleanup k8s` command to uninstall all Inspect Helm charts.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add `unset` magic string for `runtimeClassName` which will remove the field from the pod spec.
 - Add ignored `timeout_retry` parameter to `exec()` method.
 - Always capture the output of `helm uninstall` so that errors can contain meaningful information.
 - Add support for `inspect sandbox cleanup k8s` command to uninstall all Inspect Helm charts.
--- a/docs/docs/helm/built-in-chart.md
+++ b/docs/docs/helm/built-in-chart.md
@@ -14,10 +14,11 @@ familiar with them.
 In addition to the info below, see `agent-env/README.md` and `agent-env/values.yaml` for
 a full list of configurable options.
 
-## Container runtime (gVisor)
+## Container runtime class (gVisor)
 
 The default container runtime class name for every service is `gvisor` which, (depending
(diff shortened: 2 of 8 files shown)


docs/docs/helm/built-in-chart.md
```diff
@@ -27,11 +28,55 @@ services:
...
+    runc     runc      42d
+    ```
+
+??? question "Aren't containerd or CRI-O the container runtimes?"
+
+    There are multiple "levels" of container runtime in Kubernetes. containerd or CRI-O
+    are the "high level" CRI implementations which Kubernetes uses to manage containers.
+    The discussion in this section concerning `runtimeClassName` field on Pod spec is
+    about the "lower level" OCI runtimes (like `runc` or `runsc`) which are used to
+    actually run the container processes.
```
craigwalton-dsit, line 79: Maybe it is just me, but this distinction doesn't seem to be spelled out often and has caused me a lot of confusion in the past. If someone knows of good docs for this, I'm happy to replace this with a link to them!

docs/docs/helm/built-in-chart.md
```diff
@@ -14,10 +14,11 @@ familiar with them.
 In addition to the info below, see `agent-env/README.md` and `agent-env/values.yaml` for
 a full list of configurable options.
 
-## Container runtime (gVisor)
+## Container runtime class (gVisor)
 
 The default container runtime class name for every service is `gvisor` which, (depending
-on your cluster) should map to the `runsc` runtime. You can override this if required.
+on your cluster) should map to the `runsc` runtime handler. You can override this if
+required.
```
tcatling, line 21: Is it worth being explicit that not overriding this requires a `RuntimeClass` deployed in the cluster? We could provide an example:

```
apiVersion: node.k8s.io/v1
kind: RuntimeClass
metadata:
  name: runc
handler: runc # EKS specific
```

docs/docs/helm/built-in-chart.md
tcatling: @alex-remedios-aisi might have some input but i've not seen a clearer explanation than this

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add `unset` magic string for `runtimeClassName` which will remove the field from the pod spec.
```
tcatling, line 5: I'm not sure if 'unset' is the clearest possible magic string here. How about `CLUSTER_DEFAULT` or something instead?

tcatling (approved):
This seems like a good fix to me. Left some minor comments but nothing blocking

docs/docs/helm/built-in-chart.md
craigwalton-dsit: Perhaps! I perhaps haven't done a great split, but I'd rather keep these sorts of config examples on the [remote cluster setup page](https://k8s-sandbox.ai-safety-institute.org.uk/getting-started/remote-cluster/). I've added a link to it from here. Do you think the examples there need augmenting?

tcatling: aha, that makes sense! No, all seems good if there's snippets there

PR #56, opened 2025-02-04
Prevent DNS exfiltration attacks

craigwalton-dsit:
Please see the `### DNS Exfiltration` docs section added in this PR for the motivation.

I'll advertise the following changes in behaviour to users:
1. Only DNS lookups to services part of the eval and any `allowedDomains` are permitted.
2. Reverse DNS lookups (i.e. IP address to domain name) are no longer permitted.
3. Requests to domains which aren't on the allow list will now typically fail more quickly with "temporary failure in DNS resolution" or similar rather than a long timeout (this depends on the client used though).

I've done some validation against UK AISI evals (I'll discuss this internally).

### DIFF
--- a/docs/docs/security/network-access.md
+++ b/docs/docs/security/network-access.md
@@ -4,8 +4,8 @@ It is good security practice to prevent your containers from communicating with 
 internet by default.
 
 However, some evals may require internet access (e.g. to install packages or research
-topics). The [built-in Helm chart](../helm/built-in-chart.md) allows you to specify a list
-of domains that your containers can access.
+topics). The [built-in Helm chart](../helm/built-in-chart.md) allows you to specify a
+list of domains that your containers can access.
 
 ## Cilium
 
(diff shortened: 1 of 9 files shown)


src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
```diff
@@ -25,7 +25,12 @@ spec:
               protocol: ANY
           rules:
             dns:
-              - matchPattern: "*"
+              {{- /* Prevent DNS exfiltration attacks. Only allow DNS lookups for */}}
```
craigwalton-dsit, line 28: TODO: I think this will break cases where a `helm-values.yaml` file uses

```
allowEntities:
  - "world"
```

Presumably the `allowCIDR` support doesn't matter wrt DNS because you're probably accessing an IP address directly without any DNS lookup.

test/k8s_sandbox/test_networks.py
```diff
@@ -17,7 +17,7 @@ async def sandbox() -> AsyncGenerator[K8sSandboxEnvironment, None]:
 
 
 async def can_ping_service(service: str, sandbox: K8sSandboxEnvironment) -> bool:
-    result = await sandbox.exec(["ping", "-c", "1", service], timeout=3)
+    result = await sandbox.exec(["ping", "-c", "1", "-n", service], timeout=3)
```
craigwalton-dsit, line 20: `-n` prevents reverse DNS lookups. With this new setup, reverse DNS lookups do not work, so the `ping` command ends up taking ~5s as I think it attempts reverse DNS lookups a few times.

This does not affect other commands like `wget` or `curl` - they're happy to proceed as soon as they've resolved the IP.

tcatling:
See comment [here](https://github.com/AI-Safety-Institute/aisi-research-platform/issues/356#issuecomment-2644794803)

tcatling:
> change in behaviour: requests e.g. curl or wget to blocked domains will now fail with a "temporary failure in DNS resolution" or similar, rather than a timeout

This seems like a bonus!

src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
```diff
@@ -25,7 +29,13 @@ spec:
               protocol: ANY
           rules:
             dns:
```
jasongwartz-aisi, line 31: Very cool, nice implementation!

jasongwartz-aisi (approved):
Looks good to me, great improvement! Have made one non-blocking suggestion on wording.

docs/docs/security/network-access.md
```diff
@@ -19,3 +19,26 @@ using the built-in Helm chart due to how DNS resolution is handled.
...
+See the [limitations](../design/limitations.md) section for how this may affect your
+evals.
+
+??? danger "I don't want to restrict DNS lookups"
+
+    Please consider the security implications; DNS exfiltration will be possible **and**
+    your containers will have unrestricted internet access.
+
+    To allow **all** DNS queries and **all** internet access, set
+    `allowDomains: ["*"]` or `allowEntities: "all"` in the `values.yaml` file.
```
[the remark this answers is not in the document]
craigwalton-dsit: Totally agree we should spell this out. Will update!

src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
craigwalton-dsit: Tom's idea!

jasongwartz-aisi (approved):

PR #57, opened 2025-02-04
Make examples valid `helm-values.yaml` files (by adding `services` key)

craigwalton-dsit:
Thanks for spotting this @tcatling!


tcatling (approved):
ty!

PR #58, opened 2025-02-07
Tighten Ruff rules

craigwalton-dsit:
These were copied from another repo originally. There are some ignores (E203: whitespace-before-punctuation) which we don't need to ignore.

And we _do_ want to be strict on the `D10x` (missing docstrings in public members) rules within `src` but not within `test`.

### DIFF
--- a/src/k8s_sandbox/__init__.py
+++ b/src/k8s_sandbox/__init__.py
@@ -1,3 +1,5 @@
+"""Package for a Kubernetes sandbox environment provider for Inspect AI."""
+
 from k8s_sandbox._pod import GetReturncodeError, PodError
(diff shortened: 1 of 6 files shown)


src/k8s_sandbox/__init__.py
```diff
@@ -1,3 +1,5 @@
+"""Package for a Kubernetes sandbox environment provider for Inspect AI."""
```
craigwalton-dsit, line 1: To be compliant with D104 Missing docstring in public package

jasongwartz-aisi (approved):

jasongwartz-aisi (approved):

PR #59, opened 2025-02-10
Document `inspect sandbox cleanup k8s`

craigwalton-dsit:
Was going to link to docs site on a Slack message and realised I haven't actually documented this handy command!

### DIFF
--- a/docs/docs/tips/cleanup.md
+++ b/docs/docs/tips/cleanup.md
@@ -19,5 +19,6 @@ Inspect-managed based on the labels) in the current namespace and offer to unins
 them all for you.
 
 !!! warning
-    If you are using a shared Kubernetes namespace, please be mindful of other
-    users' work.
+    This command will find and uninstall all Inspect-managed Helm releases **for any
+    user of the Kubernetes namespace**. If you are using a shared Kubernetes namespace,
+    please be careful when choosing which Helm releases to uninstall.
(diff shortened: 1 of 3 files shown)


docs/docs/tips/cleanup.md
```diff
@@ -0,0 +1,23 @@
...
+inspect sandbox cleanup k8s
+```
+
+This will list all of the Inspect-managed Helm releases (it will infer whether they are
+Inspect-managed based on the labels) in the current namespace and offer to uninstall
+them all for you.
+
+!!! warning
+    If you are using a shared Kubernetes namespace, please be mindful of other
+    users' work.
```
jasongwartz-aisi, line 23: This sounds a bit ambiguous - I would specifically mention (either here, or maybe with `all of the Inspect-managed Helm releases`) that this is all Inspect-managed Helm releases in the namespace **for any user**

Maybe something like:
```suggestion
    This command will find and uninstall all Inspect-managed Helm releases
    **for any user of the Kubernetes namespace**. If you are using a shared Kubernetes
    namespace, please be careful when choosing which Helm releases to uninstall.
```

craigwalton-dsit: I like it, thanks!

jasongwartz-aisi (approved):

PR #63, opened 2025-02-19
Handle the case where namespace isn't specified, and default to 'default'

tcatling:
'namespace' is often not explicitly included in context entries and the convention here is to target 'default'. In order to repro this, try creating a cluster with `kind` then using it to spawn a sandbox, and you'll get an exception due to the explicit lookup of a namespace.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- If a namespace is not includes in the kubeconfig context, default to a namespace named "default".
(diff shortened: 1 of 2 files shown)


craigwalton-dsit (approved):
Thanks for the fix!
I guess because within UK AISI, we automatically set the current namespace on users' instances with `kubectl config set-context --current --namespace=...` we've never run into this.
I took the liberty of adding a changelog entry.

PR #69, opened 2025-02-27
Add `karpenter.sh/do-not-disrupt: "true"`

craigwalton-dsit:
See docs https://karpenter.sh/v0.32/concepts/disruption/#controls

Selfishly, this is UK AISI infrastructure-relevant, but it will not cause any issues for clusters which don't use Karpenter.

Read more about why we want to avoid disruptions here https://k8s-sandbox.ai-safety-institute.org.uk/design/limitations/#containers-may-restart


craigwalton-dsit:
Closing. Consensus was this is too infrastructure-dependent. We'll investigate adding a cluster-wide policy within our cluster to have the same effect.

PR #78, opened 2025-03-05
Update inspect_ai, resolve type checking issue

craigwalton-dsit:
The return type of Inspect's `concurrency()` has changed, therefore mypy complains.

This only affects type-checking, doesn't have any bearing on running evals, therefore this change is forward & backward compatible.


jasongwartz-aisi (approved):
In the commit or release notes, could you link to the corresponding change in Inspect?

PR #79, opened 2025-03-05
Retry on resource quota errors

craigwalton-dsit:
**DNM** Given the below, does this PR add enough value to merit its inclusion?

This is only relevant to clusters which use resource quotas.

Note: Different resource quotas manifest exceeding them in different ways: for a `Pod` quota, the Helm release will eventually time out with no mention of quotas; for a `configmap` quota, an error matching the regex in this PR will be raised immediately.

https://github.com/user-attachments/assets/fb2de85a-70eb-4a6a-bfd5-238e57eb40b8

I've also tested this manually by lowering my namespace's `configmap` resource quota to 1 (see screen recording)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -896,13 +896,13 @@ files = [
 
 [[package]]
 name = "inspect-ai"
-version = "0.3.69"
+version = "0.3.72"
 description = "Framework for large language model evaluations"
 optional = false
 python-versions = ">=3.10"
 files = [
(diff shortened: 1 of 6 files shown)


tcatling (commented):
internal users do hit this quite a bit, almost to the extent that we're using resource quotas as a throttle instead of sandbox limits etc. Given the issues we've had with leftover deployments (although these should now be much improved) i think this would be a nice thing to surface to users

craigwalton-dsit:
Thanks, useful context! But are users hitting configmap quotas (which are higher than pod quotas for us)? Because if they're hitting pod quotas, this PR sadly won't do anything for them.

craigwalton-dsit:
Closing due to uncertainty about this actually being useful: this doesn't help if you hit Pod quotas.
Happy to reopen if there's appetite for discussion.

PR #82, opened 2025-03-06
Use Helm's `--ignore-not-found` flag when uninstalling

craigwalton-dsit:
I wasn't aware of this flag until today - it is a much more elegant solution than what I did in #67. I should have checked if there was such an option back then!

It requires Helm 3.13.0 from September 2023 ([release notes](https://github.com/helm/helm/releases/tag/v3.13.0)). Previously we required only 3.10.0 (this was a fairly arbitrary choice though). 3.17.2 is the latest at time of writing.


test/k8s_sandbox/test_helm.py
```diff
@@ -92,11 +92,15 @@ async def test_invalid_helm_timeout(
...
-        await Release(__file__).install()
+        await release.install()
 
     # Verify that we detect the install timeout and add our own message.
     assert "The configured timeout value was 1s. Please see the docs" in str(
         excinfo.value
     )
+    # The release probably won't have been installed given the short timeout, but clean
+    # up just in case.
+    await release.uninstall(quiet=True)
```
craigwalton-dsit, line 106: When running these tests manually, I noticed a release got left behind.

jasongwartz-aisi (approved):
nice improvement!

Issue #88, opened 2025-03-24
Add kubeconfig context to `K8sSandboxEnvironmentConfig`


craigwalton-dsit:
See Slack https://inspectcommunity.slack.com/archives/C0855KJ6BMW/p1742689094049239

Sami would like to be able to set the kubeconfig context on a sample/task basis.

Similar to #60

craigwalton-dsit:
Started hacking this out in `craig/kubeconfig-context`. This is a first iteration and needs more consideration (i.e. not quite ready to PR!).

craigwalton-dsit:
Used eval to manual testing

```py
from inspect_ai import Task, eval, task
from inspect_ai.dataset import Sample
from inspect_ai.scorer import includes
from inspect_ai.solver import generate, use_tools
from inspect_ai.tool import bash
from inspect_ai.util import SandboxEnvironmentSpec

from k8s_sandbox._sandbox_environment import K8sSandboxEnvironmentConfig


@task
def multiple_contexts():
    return Task(
        dataset=[
            Sample(
                input="Run `cat /etc/os-release | grep VERSION_CODENAME`. Reply with codename only.",
                target="bookworm",
                sandbox=SandboxEnvironmentSpec(
                    "k8s",
                    K8sSandboxEnvironmentConfig(context="minikube"),
                ),
            ),
            Sample(
                input="Run `cat /etc/os-release | grep VERSION_CODENAME`. Reply with codename only.",
                target="bookworm",
                sandbox=SandboxEnvironmentSpec(
                    "k8s",
                    K8sSandboxEnvironmentConfig(
                        context="arn:aws:eks:eu-west-2:302263072849:cluster/rpv2-prod"
                    ),
                ),
            ),
            Sample(
                input="Run `cat /etc/os-release | grep VERSION_CODENAME`. Reply with codename only.",
                target="bookworm",
                sandbox=SandboxEnvironmentSpec(
                    "k8s",
                    K8sSandboxEnvironmentConfig(
                        context="arn:aws:eks:us-east-1:302263072849:cluster/rpv2-prod"
                    ),
                ),
            ),
            Sample(
                input="Run `cat /etc/os-release | grep VERSION_CODENAME`. Reply with codename only.",
                target="bookworm",
                sandbox=SandboxEnvironmentSpec(
                    "k8s", K8sSandboxEnvironmentConfig(context=None)
                ),
            ),
        ],
        solver=[
[...]

PR #89, opened 2025-03-27
Add support for specifying a kubeconfig context name

craigwalton-dsit:
As requested by Sami.

This allows users to have a `Sample` target a specific cluster e.g. if a `Sample` requires GPU access and your primary cluster doesn't have suitable nodes. You can therefore be interacting with multiple clusters within the scope of a single eval / inspect process.

See #88

## Concurrency
There are 2 considerations:
1. the Python `kubernetes` library is synchronous, so we use threads to parallelise usage of it
2. the kubeconfig file can change on disk between loading it and the Helm subprocess or kubernetes client using it

I've mitigated 1 by use of locks and thread locals.

For 2, as we're "telling" Helm which context _name_ to use (rather than e.g. passing it the immutable context dict itself), we'd have to take steps like creating a copy of the kubeconfig file on disk and having Helm use that. However, I'm not convinced that the added complexity is worth it - if the user is deleting contexts from their kubeconfig while an eval which references it is running, maybe it is on them?

In other systems, I'd take a more hard line approach and insist that if a race condition can be mitigated, we do so, but I'm not sure where we want to draw the line for this package. Comments/disagreements welcome!

### DIFF
--- a/src/k8s_sandbox/_kubernetes_api.py
+++ b/src/k8s_sandbox/_kubernetes_api.py
@@ -120,7 +120,7 @@ class _Config:
                 return context
         raise ValueError(
             f"Could not find a context named '{context_name}' in the kubeconfig file. "
-            f"Available contexts: {self.contexts}."
+            f"Available contexts: {[ctx['name'] for ctx in self.contexts]}."
         )
 
 
--- a/src/k8s_sandbox/_kubernetes_api.py
+++ b/src/k8s_sandbox/_kubernetes_api.py
@@ -2,43 +2,157 @@ from __future__ import annotations
 
 import logging
 import threading
+from typing import Any
 
 from kubernetes import client, config  # type: ignore
 
 logger = logging.getLogger(__name__)
 
 _thread_local = threading.local()
-_load_config_lock = threading.Lock()
-_config_loaded = False
 
 
-def k8s_client() -> client.CoreV1Api:
+def k8s_client(context_name: str | None) -> client.CoreV1Api:
     """
-    Gets a thread-local Kubernetes client.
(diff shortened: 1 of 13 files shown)


src/k8s_sandbox/_kubernetes_api.py
```diff
@@ -2,43 +2,157 @@
...
+            raise ValueError(
+                f"Could not find a context named '{context_name}' in kubeconfig "
+                "because no contexts were present in the kubeconfig file."
+            )
+        for context in self.contexts:
+            if context["name"] == context_name:
+                return context
+        raise ValueError(
+            f"Could not find a context named '{context_name}' in the kubeconfig file. "
+            f"Available contexts: {self.contexts}."
```
jasongwartz-aisi, line 123: Should this maybe print a list of names? I assume this will be a lot of output otherwise.
Something like this? (not tested!)
```suggestion
            f"Available contexts: {",".join([context["name"] for context in self.contexts])}."
```

jasongwartz-aisi (approved):
overall lgtm, thanks for the improvement!

src/k8s_sandbox/_kubernetes_api.py
craigwalton-dsit: That would be much more helpful! Actioned. Sample error:
```
ValueError("Could not find a context named 'does-not-exist' in the kubeconfig file. Available contexts: ['arn:aws:eks:eu-west-2:302263072849:cluster/rpv2-prod', 'arn:aws:eks:us-east-1:302263072849:cluster/rpv2-prod', 'minikube'].")
```

jasongwartz-aisi (approved):

PR #92, opened 2025-04-07
Implement `SandboxConnection`

craigwalton-dsit:
This is useful for human agent baselining or for providing users with a command to copy/paste to get a shell into the container.

This is not aimed at preventing METR implementing a method for doing this over SSH (as their human baseliners won't have cluster credentials) - it is simply aimed at ensuring the default method of creating a `SandboxConnection` is the vanilla `kubectl` option. We can definitely add alternative methods in future!

It is a bit janky as an error pops up despite command working - this may just be related to how we're using VS Code at AISI though (via SSH to EC2 instances).

Using Terminal:

https://github.com/user-attachments/assets/e9a2ff0c-1cca-47d1-95ee-f9e209b7821b

Using Window:
Sadly there's no way to specify the container name to connect to, or the kubeconfig context when using `remote-containers.attachToK8sContainer`. No docs or easy to access source code for this easy AFAICT.

https://github.com/user-attachments/assets/67c7a650-c14c-4e71-a57f-1590f284e8af

### DIFF
--- a/src/k8s_sandbox/_sandbox_environment.py
+++ b/src/k8s_sandbox/_sandbox_environment.py
@@ -1,3 +1,4 @@
+import shlex
 import tempfile
 from contextlib import contextmanager
 from pathlib import Path
@@ -6,6 +7,7 @@ from typing import Any, Generator, Literal, cast, overload
 from inspect_ai.util import (
     ExecResult,
     OutputLimitExceededError,
+    SandboxConnection,
     SandboxEnvironment,
     SandboxEnvironmentConfigType,
     sandboxenv,
@@ -186,6 +188,29 @@ class K8sSandboxEnvironment(SandboxEnvironment):
                 return (
                     temp_file.read() if not text else temp_file.read().decode("utf-8")
                 )
+
+    async def connection(self) -> SandboxConnection:
+        pod = self._pod.info
+        kubectl_cmd = [
+            "kubectl",
+            "exec",
+            "-it",
+            pod.name,
+            "-n",
+            pod.namespace,
+            "-c",
+            pod.default_container_name,
+        ]
+        if pod.context_name is not None:
+            kubectl_cmd.extend(["--context", pod.context_name])
+        kubectl_cmd.extend(["--", "bash", "-l"])
+        return SandboxConnection(
+            type="k8s",
+            command=shlex.join(kubectl_cmd),
+            # TODO: Can we use `remote-containers.attachToK8sContainer`?
+            vscode_command=None,
+            container=pod.default_container_name,
+        )
 
     @contextmanager
     def _log_op(
--- a/src/k8s_sandbox/_sandbox_environment.py
+++ b/src/k8s_sandbox/_sandbox_environment.py
@@ -207,8 +207,15 @@ class K8sSandboxEnvironment(SandboxEnvironment):
         return SandboxConnection(
             type="k8s",
             command=shlex.join(kubectl_cmd),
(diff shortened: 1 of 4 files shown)


src/k8s_sandbox/_sandbox_environment.py
```diff
@@ -187,6 +189,36 @@ async def read_file(self, file: str, text: bool = True) -> str | bytes:
...
+        if pod.context_name is not None:
+            kubectl_cmd.extend(["--context", pod.context_name])
+        kubectl_cmd.extend(["--", "bash", "-l"])
+        return SandboxConnection(
+            type="k8s",
+            command=shlex.join(kubectl_cmd),
+            # Note that there is no facility to specify the kubeconfig context or the
+            # default container name.
+            vscode_command=[
+                "remote-containers.attachToK8sContainer",
```
tcatling, line 213: what happens here if the extension is not installed?

craigwalton-dsit: Good point. I've just tried this (both by disabling the extension and by using a fake extension) and it appears to fail silently.

Maybe there is something Inspect could do to check if the command was executed successfully, but the integration with VS Code doesn't seem to have a really nice API (it writes a file with the command encoded in json into some directory which I think vs code magically picks up...) https://github.com/UKGovernmentBEIS/inspect_ai/blob/50e120d1dbc531bf50c51474094aeed7fc5a96d3/src/inspect_ai/_util/vscode.py#L23

I don't think that the k8s_sandbox package at this point in execution should be responsible for checking what is installed as:
1. this code will get executed regardless of if the user even has VS Code installed/is wanting to use the VS Code functionality
2. This method, while named `connection()` is more of a `get_connection_info()` - it doesn't actually execute anything - it is up to Inspect to act upon that info if applicable IMO.


Min repro:
```py
from inspect_ai._util.vscode import VSCodeCommand, execute_vscode_commands

cmd = VSCodeCommand(
    command="fake.fake",
    args=[
        {
            "name": "agent-env-evtjjbcx-default-0",
            "namespace": "craig-walton-default",
        }
    ],
)

execute_vscode_commands(cmd)
```

src/k8s_sandbox/_sandbox_environment.py
```diff
@@ -187,6 +189,36 @@ async def read_file(self, file: str, text: bool = True) -> str | bytes:
...
+            pod.namespace,
+            "-c",
+            pod.default_container_name,
+        ]
+        if pod.context_name is not None:
+            kubectl_cmd.extend(["--context", pod.context_name])
+        kubectl_cmd.extend(["--", "bash", "-l"])
+        return SandboxConnection(
+            type="k8s",
+            command=shlex.join(kubectl_cmd),
```
tcatling, line 209: are 'command' vs. 'vscode_command' presented as 'VS Code Terminal' and 'VS Code Windows' in the gif you attached? What will this look like if I'm just running inspect in a bare terminal outside of vscode?

craigwalton-dsit: Yep exactly. You'd see just:

Screenshot 2025-04-07 at 12 12 43


Note even in VS Code, it always prints out the command you can copy/paste manually:

Screenshot 2025-04-07 at 12 12 33

tcatling: nifty!

tcatling (approved):
this looks great to me, super nice to support this feature

Issue #91, opened 2025-04-03
Allow configuring ingress ports


tbroadley:
At METR, we allow human baseliners to access sandbox environments over SSH. To support this, we'd like to make it possible to configure specific ports on which Cilium should allow ingress to the default sandbox environment.

This would look like 

- Adding a new key (`externalIngressPorts`?) to `values.yaml`, which would take a list of port numbers
- Adding a new CiliumNetworkPolicy to `network-policy.yaml` that would allow ingress on the specified ports

craigwalton-dsit:
Closing this as it can be done via #93 now.

PR #93, opened 2025-04-10
Add support for user-provided network policies via `additionalResources` templating

craigwalton-dsit:
This is to support METR's request of allowing SSH ingress to the default service for the purpose of human baselining (#91). This is where a human attempts to perform the task.

This can be done by passing `additionalResources` to the Helm templating engine (see docs example).

~I've opted to keep this simple with an array of integers, rather than say an array of objects containing a port number and a protocol. I allow `protocol: ANY` for the listed port numbers.~

~While the requirement is just to allow ingress to the default service, I've put the new `externalIngressPorts ` field at the service level (as opposed to top level) because this makes it self-documenting as to which services this applies to (and as a nice side effect, allows you to optionally allow ingress to non-default services).~


craigwalton-dsit:
**After internal discussion**
Is ingress into evals something we want to support 1st class? If it is required for an eval (say a honeypot which accepts real internet traffic) then maybe. But if it is for supporting the running of evals, maybe not. Is there a way in which we could support this via the `additionalResources` if we make it work with basic templating? I don't think so as we'd need stuff like `matchLabels: {{- include "agentEnv.selectorLabels" $ | nindent 6 }}`. So we'd need to feed the contents of `additionalResources` through the templating engine.

E.g. you'd add something like this to `helm-values.yaml`:
```yaml
apiVersion: cilium.io/v2
kind: CiliumNetworkPolicy
metadata:
  name: {{ template "agentEnv.fullname" $ }}-sandbox-default-external-ingress
  annotations:
    {{- toYaml $.Values.annotations | nindent 4 }}
spec:
  description: Allow external ingress from all entities to the default service.
  endpointSelector:
    matchLabels:
      io.kubernetes.pod.namespace: {{ $.Release.Namespace }}
      {{- include "agentEnv.selectorLabels" $ | nindent 6 }}
      inspect/service: default
  ingress:
    - fromEntities:
      - all
      toPorts:
      - ports:
        - port: 2222
          protocol: TCP
```

and we'd do something like

```
{{- if .Values.additionalResources }}
{{- range $resource := .Values.additionalResources }}
---
{{ tpl (toYaml $resource) $ }}
{{- end }}
{{- end }}
```

docs/docs/helm/built-in-chart.md
```diff
@@ -384,6 +384,30 @@ annotations:
...
+
+```yaml
+services:
+  default:
+    externalIngressPorts:
+      - 2222
+```
+
+This will create an additional Cilium Network Policy to allow ingress on the specified
+ports over any protocol from all entities (see [Cilium
```
tcatling, line 400: maybe it would be better to add something like `extraEgressRules` and `extraIngressRules`? I can imagine people wanting to manage ingress from certain CIDR blocks, endpoints etc rather than everywhere, and this would provide a reasonable future-proof escape hatch

craigwalton-dsit: Good idea. So our template would be along the lines of

```yaml
{{- range $name, $service := .Values.services }}
{{- if $service.extraIngressRules }}
---
apiVersion: cilium.io/v2
kind: CiliumNetworkPolicy
metadata:
  name: {{ template "agentEnv.fullname" $ }}-sandbox-{{ $name }}-extra-ingress
  annotations:
    {{- toYaml $.Values.annotations | nindent 4 }}
spec:
  description: |
    Allow additional ingress to the "{{ $name }}" service.
  endpointSelector:
    matchLabels:
      io.kubernetes.pod.namespace: {{ $.Release.Namespace }}
      {{- include "agentEnv.selectorLabels" $ | nindent 6 }}
      inspect/service: {{ $name }}
  ingress:
    {{ $service.extraIngressRules | toYaml }}
{{- end }}
{{- end }}
```

And the user could define in their values.yaml:

```yaml
services:
  default:
    extraIngressRules:
    - fromEntities:
      - all
      toPorts:
      - ports:
        - port: 2222
          protocol: TCP
```
?

tcatling: yes exactly! the more stuff we can defer to `| toYaml }` the better, imo. Schema validation can often be difficult when doing this, but if you want to enforce it strictly you could do something like [this](https://stackoverflow.com/questions/71089827/is-there-any-easy-way-to-convert-crd-customresourcedefinition-to-json-schema) to create schemas for ingress/egress entries.

craigwalton-dsit: Nice, the only thing I wonder is whether need need/want to split `extraEgressRules` and `extraIngressRules` or whether we just have 1 `extraNetworkPolicy` like

```yaml
services:
  default:
    extraNetworkPolicy:
    ingress:
      - fromEntities:
        - all
        toPorts:
        - ports:
          - port: 2222
            protocol: TCP
```
?

tcatling: Actually i think the most flexible and idiomatic approach here is probably to accept `extraResources` and just dump the entirety into `extra-resources.yaml`. This allows people to include any extra stuff they need. If we document an example of using this to add an extra cilium ingress policy, i think that should be sufficient? would need to ensure that templating still works within these extra snippets

test/k8s_sandbox/helm_chart/resources/additional-resources-template-block-values.yaml
```diff
@@ -0,0 +1,23 @@
+additionalResources:
+- |
+  apiVersion: cilium.io/v2
+  kind: CiliumNetworkPolicy
+  metadata:
+    name: {{ template "agentEnv.fullname" $ }}-sandbox-default-external-ingress
+    annotations:
+      {{- toYaml $.Values.annotations | nindent 6 }}
```
craigwalton-dsit, line 8: Lines like this are not valid standalone YAML, hence why this whole resource is in a string block.

PR #94, opened 2025-04-15
Support `user` parameter on `SandboxEnvironment.exec()`

craigwalton-dsit:
Use `runuser` to be able to `exec` commands as different users.

If any of the following occur, an error is raised which will cause the eval to fail (not just a failed ExecResult). The error message will include a link to the relevant docs.
- container not running as root
- container doesn't have `runuser`
- specified user doesn't exist

This is intentional as it is most likely a programming error/bug, not an LLM feeding in something silly.

See comment below on Inspect/Platform Infra discussion on this.

### DIFF
--- a/src/k8s_sandbox/_pod/execute.py
+++ b/src/k8s_sandbox/_pod/execute.py
@@ -9,11 +9,13 @@ from inspect_ai.util import SandboxEnvironmentLimits as limits
 from kubernetes.stream.ws_client import WSClient  # type: ignore
 
 from k8s_sandbox._pod.buffer import LimitedBuffer
+from k8s_sandbox._pod.error import ExecutableNotFoundError
 from k8s_sandbox._pod.get_returncode import get_returncode
 from k8s_sandbox._pod.op import PodOperation
 
 COMPLETED_SENTINEL = "completed-sentinel-value"
 COMPLETED_SENTINEL_PATTERN = re.compile(rf"<{COMPLETED_SENTINEL}-(\d+)>")
+EXEC_USER_URL = "https://k8s-sandbox.aisi.org.uk/design/limitations#exec-user"
 
 
 class ExecuteOperation(PodOperation):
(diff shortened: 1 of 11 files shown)


test/k8s_sandbox/test_inspect_self_check.py
```diff
@@ -28,8 +28,8 @@ async def non_root(
 
 async def test_self_check_k8s_default_root(root: SandboxEnvironment) -> None:
     known_failures = [
-        # The user parameter is not supported in K8sSandboxEnvironment.
-        "test_exec_as_user",
+        # Running as a nonexistent user results in an exception being raised (by design)
+        # whereas the test expects a ExecResult to be returned without an exception.
```
art-dsit, line 32: I think raising an exception is better than how the docker sandbox does it - let's change the signature of `exec` at some future point to define the exception thrown if the user doesn't exist

craigwalton-dsit:
As discussed, let's explain that this isn't recommended on the docs, and link to the docs from the error messages

src/k8s_sandbox/_pod/execute.py
```diff
@@ -23,26 +25,41 @@ def exec(
...
-            stderr=True,
-            stdin=True,
-            stdout=True,
-            # Leave stdout and stderr as binary. Has no effect on stdin.
-            binary=True,
-        )
+    def _interactive_shell(self, user: str | None) -> Generator[WSClient, None, None]:
+        command = ["/bin/sh"]
+        if user is not None:
+            command = ["runuser", "-u", user] + command
```
craigwalton-dsit, line 43: The `runuser` is intentionally injected at this top level, not within the shell script which wraps the user command.

This is so that all other aspects of the `exec()` are faithful to the relevant user's permissions e.g. changing directory. This also minimises the risk of a malicious user command somehow breaking out of the `runuser` (e.g. exploiting the shell script we generate).

test/k8s_sandbox/test_inspect_self_check.py
craigwalton-dsit: Sounds good!

tcatling:
Does this work in a container where the default user isn't root or sudo-enabled? The top level `runuser` will run as the default container user, which needs enough privileges to do the switching. 

I can imagine confusing situations when trying to execute stuff in a container which already has a non-privileged user, e.g. `nginx`, but if you actually specify `user="nginx"` in the exec call, it will fail because of lack of permissions 🤔

tcatling (approved):
this all lgtm. I agree that raising an exception on any of the conditions you list seems sensible

src/k8s_sandbox/_pod/execute.py
```diff
@@ -23,26 +25,41 @@ def exec(
...
+                stderr=True,
+                stdin=True,
+                stdout=True,
+                # Leave stdout and stderr as binary. Has no effect on stdin.
+                binary=True,
+            )
+        # Raised if /bin/sh or runuser cannot be found in the Pod (not if a
+        # user-supplied) command cannot be found.
+        except ExecutableNotFoundError as e:
+            if 'error finding executable "runuser"' in str(e):
```
art-dsit, line 56: This if statement seems redundant - this is an internal Error class, so we know all the places in which it's created

craigwalton-dsit: I agree that we know when it is raised - but it could be raised if `/bin/sh` wasn't found so I thought we'd only want to show the user the educational message if this is specifically that `runuser` wasn't found. If it that `/bin/sh` wasn't found, we let that exception propagate. Lmk if I've misunderstood your suggestion!

art-dsit: Ah I see, I didn't read the comment about /bin/sh. Although I'd say that If /bin/sh isn't found, seems unlikely we would get this far as nothing would work. But maybe one day when we have Windows pods.

art-dsit (approved):

PR #95, opened 2025-04-15
Support `user` parameter on `SandboxEnvironment.connection()`

craigwalton-dsit:
Not to be confused with similar #94 

`SandboxEnvironment.connection()` returns commands which the user can use to connect to a sandbox. Inspect PR https://github.com/UKGovernmentBEIS/inspect_ai/pull/1658 now adds an optional user parameter to this method.

### DIFF
--- a/src/k8s_sandbox/_sandbox_environment.py
+++ b/src/k8s_sandbox/_sandbox_environment.py
@@ -189,34 +189,12 @@ class K8sSandboxEnvironment(SandboxEnvironment):
                     temp_file.read() if not text else temp_file.read().decode("utf-8")
                 )
 
-    async def connection(self) -> SandboxConnection:
-        pod = self._pod.info
-        kubectl_cmd = [
-            "kubectl",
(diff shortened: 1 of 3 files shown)


src/k8s_sandbox/_sandbox_environment.py
```diff
@@ -254,6 +232,42 @@ def _log_op(
...
+            "-n",
+            self._pod.info.namespace,
+            "-c",
+            self._pod.info.default_container_name,
+        ]
+        if self._pod.info.context_name is not None:
+            kubectl_cmd.extend(["--context", self._pod.info.context_name])
+        kubectl_cmd.append("--")
+        if user is not None:
+            kubectl_cmd.extend(["su", "-", user, "-s", "/bin/bash"])
```
craigwalton-dsit, line 250: Should `-s /bin/bash` be before the `-`?

tcatling (approved):
lgtm!

PR #98, opened 2025-05-13
Support fractional byte values

rasmusfaber:
Docker compose byte values (i.e. memory) are allowed to be a fractional value. Fortunately, so are Kubernetes `resource.Quantity`s.

This PR just adds this to the regexp in `_converter.convert_unit()` and adds a single test.


craigwalton-dsit (approved):
Thanks Rasmus! This is a great improvement. I think this clearly falls within the scope of this compose->helm feature as described here https://k8s-sandbox.aisi.org.uk/helm/compose-to-helm/#limitations

PR #104, opened 2025-05-21
Ignore `x-local` key when converting from Compose to Helm values

craigwalton-dsit:
See discussion here re swebench https://github.com/UKGovernmentBEIS/inspect_evals/issues/212#issuecomment-2896690871

From the Inspect docs:
>The ctf-agent-environment is not an image that exists on a remote registry, so we add the x-local: true to indicate that it should not be pulled. If local images are tagged, they also will not be pulled by default (so x-local: true is not required).

For k8s, we always pull images. We'll emit a warning if `x-local: true`, and silently ignore it if it is any other value.

### DIFF
--- a/test/k8s_sandbox/compose/test_converter.py
+++ b/test/k8s_sandbox/compose/test_converter.py
@@ -677,7 +677,9 @@ services:
     result = convert_compose_to_helm_values(compose_path)
 
     assert "my-service" in result["services"]
+    assert "x-local" not in result["services"]["my-service"]
     assert "my-service-2" in result["services"]
+    assert "x-local" not in result["services"]["my-service-2"]
 
 
 def test_rejects_unsupported_service_key(tmp_compose: TmpComposeFixture) -> None:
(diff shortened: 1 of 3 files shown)


test/k8s_sandbox/compose/test_converter.py
```diff
@@ -663,6 +663,23 @@ def test_ignores_expose_key(tmp_compose: TmpComposeFixture) -> None:
...
+    x-local: true
+  my-service-2:
+    image: my-image
+    x-local: false
+""")
+
+    result = convert_compose_to_helm_values(compose_path)
+
+    assert "my-service" in result["services"]
+    assert "my-service-2" in result["services"]
```
jasongwartz-aisi, line 680: Is it worth asserting that `"x-local" not in result["services"]["my-service"]`?

craigwalton-dsit: Good idea! Updated.

jasongwartz-aisi (approved):
✅

PR #105, opened 2025-05-21
Support adding custom labels to resources

tbroadley:
In addition to being able to add custom annotations.

This PR adds a `labels` top-level key to the `values.yaml` schema, similar to the existing `annotations` key. inspect_k8s_sandbox applies these labels to all the k8s resources it creates

### DIFF
--- a/docs/docs/helm/custom-chart.md
+++ b/docs/docs/helm/custom-chart.md
@@ -10,6 +10,8 @@ the built-in `agent-env` chart.
   by Inspect.
 * Accept a top-level `annotations` object in its values schema (will typically include
   `annotations.inspectTaskName`). Apply these annotations to Pods and other resources.
+* Accept a top-level `labels` object in its values schema. Apply these labels to Pods
+  and other resources.
 * In the metadata section of each Pod, set the `app.kubernetes.io/instance` label to the
   name of the Helm release i.e. `.Release.Name`. This is an 8 character string supplied
   by the `k8s_sandbox` package. This is used by the `k8s_sandbox` package for
@@ -31,6 +33,7 @@ kind: Pod
 metadata:
   name: my-custom-chart-pod-{{ .Release.Name }}
   labels:
+    {{- toYaml $.Values.labels | nindent 4 }}
     app.kubernetes.io/name: {{ .Chart.Name }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     inspect/service: default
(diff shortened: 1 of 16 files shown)


docs/docs/helm/custom-chart.md
```diff
@@ -10,6 +10,8 @@ the built-in `agent-env` chart.
   by Inspect.
 * Accept a top-level `annotations` object in its values schema (will typically include
   `annotations.inspectTaskName`). Apply these annotations to Pods and other resources.
+* Accept a top-level `labels` object in its values schema. Apply these labels to Pods
```
craigwalton-dsit, line 13: I don't think your proposal actually requires adding a new custom chart requirement, which is a good thing, as if it did, this would constitute a breaking change (some of our evals use custom charts and their schemas probably don't accept a top-level labels property).

The reason that the top level `annotations` is mentioned in this file is because the Python code actually injects some annotations (namely the inspect task name).

So I think you can revert the changes to this file.

test/k8s_sandbox/inspect_integration/custom_chart/my-custom-chart/templates/template.yaml
```diff
@@ -3,6 +3,7 @@ kind: Pod
 metadata:
   name: my-custom-chart-pod
   labels:
+    {{- toYaml $.Values.labels | nindent 4 }}
```
craigwalton-dsit, line 6: Likewise, this can be reverted as we're not adding any new requirements to custom charts.

src/k8s_sandbox/resources/helm/agent-env/templates/services.yaml
```diff
@@ -4,6 +4,7 @@ kind: StatefulSet
 metadata:
   name: {{ template "agentEnv.fullname" $ -}}-{{ $name }}
   labels:
+    {{- toYaml $.Values.labels | nindent 4 }}
```
craigwalton-dsit, line 7: I think the places in this file where you've added labels on top of existing labels may be failing. I think I repro'ed with 
`helm template foo src/k8s_sandbox/resources/helm/agent-env/`

craigwalton-dsit (changes requested):
Hi Thomas, thanks for this contribution! In principle, I'm happy with the proposed changes. Just a couple of things to address.

tbroadley:
@craigwalton-dsit Thank you for reviewing this! I think I've addressed your comments.

craigwalton-dsit (approved):
Thanks Thomas, this looks good! And nice one figuring out how to combine the "hard-coded" labels with the values-provided-labels. One nit pick for a var name in the tests, but other than that I'm happy to merge.

PR #106, opened 2025-05-29
Compose converter: ignore and warn on `build` option

sjawhar:
I don't think we need to always error out when there's a `build` key in the compose file. Perfectly reasonable to have it there for centralizing service run and build logic while giving the option that it should be ignore when running with k8s. I considered a more generic `ignore_unsupported_keys` option but that seemed like overkill for now.

### DIFF
--- a/src/k8s_sandbox/compose/_converter.py
+++ b/src/k8s_sandbox/compose/_converter.py
@@ -21,9 +21,7 @@ class ComposeConverterError(Exception):
     pass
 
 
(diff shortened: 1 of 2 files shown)


sjawhar:
Now i'm thinking I should have just gone with `ignore_unspported_keys` after all. Or maybe `unsupported_key_action: Literal["ignore", "warn", "raise"] = "raise"`

sjawhar:
@craigwalton-dsit which way do you think we should go with this?

PR #107, opened 2025-06-09
Add missing `context_name` arg, prevent similar issues in future

craigwalton-dsit:
Disappointing bug which got through to production, noticed by Alex. Was a missing required `context_name=None` argument in a call to `uninstall()`.

My first reaction is "mypy should have caught this". But it wasn't because
a) the return type of the function containing the error was not defined (should have been `-> None`)
b) we don't run mypy with `--strict` or have `check_untyped_defs` set

I've resolved both in this PR.

Sadly tests do not cover this precise code path (they do cover other parts of the function though). This is because such a test would involve uninstalling all Helm releases in a namespace, and a user can run tests against the production cluster so I thought it would be unwise to actually try to perform this action just by running `pytest`. In a more mature system, I'd add dependency injection to allow us to mock a `helm uninstall`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,10 @@ ignore = [
 convention = "google"
 
 
+[tool.mypy]
+check_untyped_defs = true
+
+
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 markers = [
--- a/src/k8s_sandbox/_manager.py
+++ b/src/k8s_sandbox/_manager.py
@@ -146,7 +146,10 @@ async def uninstall_all_unmanaged_releases() -> None:
     ):
         print("Cancelled.")
         return
-    tasks = [helm_uninstall(release, namespace, quiet=False) for release in releases]
+    tasks = [
+        helm_uninstall(release, namespace, context_name=None, quiet=False)
+        for release in releases
+    ]
     await asyncio.gather(*tasks, return_exceptions=True)
     print("Complete.")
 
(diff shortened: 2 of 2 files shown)


src/k8s_sandbox/_manager.py
```diff
@@ -146,7 +146,10 @@ def _print_table(releases: list[str]) -> None:
     ):
         print("Cancelled.")
         return
-    tasks = [helm_uninstall(release, namespace, quiet=False) for release in releases]
+    tasks = [
+        helm_uninstall(release, namespace, context_name=None, quiet=False)
```
craigwalton-dsit, line 150: Context name is None as documented here https://k8s-sandbox.aisi.org.uk/tips/configuration/#targeting-specific-or-multiple-kubeconfig-contexts

pyproject.toml
```diff
@@ -52,6 +52,10 @@ ignore = [
 convention = "google"
 
 
+[tool.mypy]
+check_untyped_defs = true
```
craigwalton-dsit, line 56: >Type-checks the interior of functions without type annotations.

src/k8s_sandbox/_manager.py
jasongwartz-aisi: 👍  got it!

craigwalton-dsit: Great, thanks for reviewing and checking!

PR #114, opened 2025-07-31
Allow configuring coredns image and command

sjawhar:
We ran into issues with getting rate limited by docker hub when trying to pull the coredns image as our cluster scaled up real fast. One way around this is to authenticate image pulls, getting unlimited rate limit. There are Reasons why that's a bit more complicated for us than just adding an image pull secret, and IMO one should be able to configure the coredns image being used anyway (e.g. pull your own hardened image from your internal registry), so I added that functionality.


tcatling (approved):
this makes sense to me - ty!

PR #117, opened 2025-08-07
Check for container restarts

rasmusfaber:
When we run samples using k8s-sandbox, the agent can exceed the memory limits and force the container to restart. There is a [section about this](https://k8s-sandbox.aisi.org.uk/design/limitations/) in the docs. The agent can't detect the restart, so it gets confused that the sandbox has been reset. ([Slack](https://inspectcommunity.slack.com/archives/C0855KJ6BMW/p1754463522997109))

This PR makes the framework track the pod UID and the number of times the default container has been restarted.

It then validates these values before performing any operations, and either warns or raises an error if it detects a change.

### DIFF
--- a/src/k8s_sandbox/_pod/op.py
+++ b/src/k8s_sandbox/_pod/op.py
@@ -1,7 +1,7 @@
 import logging
 from abc import ABC
 from dataclasses import dataclass
-from typing import Generator
+from typing import Generator, Literal
 
 from kubernetes.stream import stream  # type: ignore
 from kubernetes.stream.ws_client import WSClient  # type: ignore
@@ -30,6 +30,9 @@ class PodInfo:
     context_name: str | None
     """The name of the kubeconfig context. If None, use the current context."""
     default_container_name: str
+    uid: str
+    initial_restart_count: int
+    restarted_container_behavior: Literal["warn", "raise"]
 
 
 class PodOperation(ABC):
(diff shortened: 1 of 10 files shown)


src/k8s_sandbox/_pod/op.py
```diff
@@ -94,6 +97,53 @@ def _discard_duplicate_channel(self, ws_client: WSClient) -> None:
...
         ws_client._all = _IgnoredIO()
 
+    def _check_for_pod_restart(self):
+        client = k8s_client(self._pod.context_name)
+        pod = client.read_namespaced_pod(
+            name=self._pod.name, namespace=self._pod.namespace
+        )
+        if pod.metadata.uid != self._pod.uid:
+            message = (
+                f"Pod UID mismatch: expected {self._pod.uid}, got {pod.metadata.uid}"
```
tcatling, line 107: I guess this indicates that an entirely new Pod has been created by the controller, for whatever reason?

rasmusfaber: Yes.

Good idea with the pod name. I have added that.

sjawhar:
Is this going to be merged?

jon-aisi (approved):
LGTM!

Couple of non-blocking concerns:
1. This approach requires remembering to add `self._check_for_pod_restart()` in every `PodOperation` subclass's main function
    - Would be great if this happened automatically, but doesn't seem like we have the existing structure to do that in the `PodOperation` itself
    - me & the AIs had a quick look for hooks in the kubernetes python client but couldn't find anything obvious
    - worst case is degraded error messaging which is not too big a deal
2. The tests only check the `exec` command, but `_check_for_pod_restart` is on all `PodOperation`s

Going to merge now though given the time this has been open

PR #122, opened 2025-09-12
Limit ingress on a service to ports specified in helm-values

willpayne23:
Limits ingress on a service to ports specified in helm-values like so:
```
  victim:
    image: path-to-victim-image:latest
    ports:
    - protocol: TCP
      port: 5800
```

If no ports are specified all ports are open (this is to maintain status quo)
If networks are specified alongside open ports then ports are only open to services on the same network and services with no networks have no ports exposed to any other service (regardless of what configuration is in the ports property)


willpayne23:
Looks like the helm install timed out for one of the new tests. It's the same configuration as the other test just without a network and was working on my machine so I suspect it's a one off? I'd rerun the action but I don't see the option to do that so I probably don't have permissions

PR #167, opened 2026-02-25
Fix singleton leak in test_kubernetes_api.py

StuartJenningsAISI:
## Summary

Depends on and includes all changes from #159.

Adds one additional commit to fix the pytest failure in that PR:

- Replace `setup_method` with an `autouse` pytest fixture that resets the `_Config` singleton both before **and after** each test
- Without the post-test reset, the last test in `TestGetDefaultNamespace` leaves a singleton populated with mocked namespace data (`"my-namespace"`), which leaks into downstream integration tests (e.g. `test_network_policy::test_allowed_fqdn`) that rely on the real kubeconfig

This is the fix identified by @sjawhar in #159, who was unable to push due to an SSO token issue.

## Test plan

- [ ] CI `test / pytest` should now pass (specifically `test_network_policy.py::test_allowed_fqdn` should no longer fail)
- [ ] All 7 tests in `test_kubernetes_api.py` should continue to pass


art-dsit:
Fixed by #159 being merged

PR #175, opened 2026-03-19
Run "not req_k8s" tests separately

bmillwood-aisi:
This both avoids running them twice and tests that they run without a k8s cluster.

(They seem not to!)

### DIFF
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -51,7 +51,10 @@ cilium install
 cilium status --wait
 cilium hubble enable --ui
 
-fi # CLUSTER
+else
+echo "Skipping cluster setup (--no-cluster)"
+fi
 
+# Everything below runs regardless of --no-cluster.
 echo "Installing uv environment..."
 uv sync --extra dev
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -9,14 +9,13 @@ for arg in "$@"; do
 done
 
 if [ "$CLUSTER" = true ]; then
+  echo "Setting up Minikube..."
+  minikube delete
+  # github actions runner has 2 cpus, 8G memory
+  minikube start --addons=gvisor --cni bridge --container-runtime=containerd --memory=4g
 
-echo "Setting up Minikube..."
-minikube delete
-# github actions runner has 2 cpus, 8G memory
-minikube start --addons=gvisor --cni bridge --container-runtime=containerd --memory=4g
-
-# Add the containerd RuntimeClass to the cluster.
-kubectl apply -f - <<EOF
+  # Add the containerd RuntimeClass to the cluster.
+  kubectl apply -f - <<EOF
 apiVersion: node.k8s.io/v1
 kind: RuntimeClass
 metadata:
@@ -24,8 +23,8 @@ metadata:
 handler: runc
 EOF
 
-# Add a mocked nfs-csi StorageClass which uses the hostpath provisioner.
-kubectl apply -f - <<EOF
+  # Add a mocked nfs-csi StorageClass which uses the hostpath provisioner.
+  kubectl apply -f - <<EOF
 apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
@@ -35,24 +34,23 @@ reclaimPolicy: Delete
 volumeBindingMode: Immediate
 EOF
 
-# Install Cilium CLI. Previously, the ghcr.io/audacioustux/devcontainers/cilium:1
-# devcontainer feature was used, but it doesn't allow a specific version of the CLI to
-# be installed and the latest version failed at `cilium install`.
-CILIUM_CLI_VERSION=v0.18.8
-CILIUM_CLI_ARCH=amd64
-echo "Installing Cilium CLI $CILIUM_CLI_VERSION $CILIUM_CLI_ARCH..."
-curl -L --fail --remote-name-all https://github.com/cilium/cilium-cli/releases/download/${CILIUM_CLI_VERSION}/cilium-linux-${CILIUM_CLI_ARCH}.tar.gz{,.sha256sum}
-sha256sum --check cilium-linux-${CILIUM_CLI_ARCH}.tar.gz.sha256sum
-sudo tar xzvfC cilium-linux-${CILIUM_CLI_ARCH}.tar.gz /usr/local/bin
-rm cilium-linux-${CILIUM_CLI_ARCH}.tar.gz{,.sha256sum}
+  # Install Cilium CLI. Previously, the ghcr.io/audacioustux/devcontainers/cilium:1
+  # devcontainer feature was used, but it doesn't allow a specific version of the CLI to
+  # be installed and the latest version failed at `cilium install`.
+  CILIUM_CLI_VERSION=v0.18.8
+  CILIUM_CLI_ARCH=amd64
+  echo "Installing Cilium CLI $CILIUM_CLI_VERSION $CILIUM_CLI_ARCH..."
+  curl -L --fail --remote-name-all https://github.com/cilium/cilium-cli/releases/download/${CILIUM_CLI_VERSION}/cilium-linux-${CILIUM_CLI_ARCH}.tar.gz{,.sha256sum}
+  sha256sum --check cilium-linux-${CILIUM_CLI_ARCH}.tar.gz.sha256sum
+  sudo tar xzvfC cilium-linux-${CILIUM_CLI_ARCH}.tar.gz /usr/local/bin
+  rm cilium-linux-${CILIUM_CLI_ARCH}.tar.gz{,.sha256sum}
 
-echo "Installing Cilium..."
-cilium install
-cilium status --wait
-cilium hubble enable --ui
-
+  echo "Installing Cilium..."
+  cilium install
+  cilium status --wait
+  cilium hubble enable --ui
 else
-echo "Skipping cluster setup (--no-cluster)"
+  echo "Skipping cluster setup (--no-cluster)"
 fi
 
 # Everything below runs regardless of --no-cluster.
--- a/.devcontainer/post-create.sh
+++ b/.devcontainer/post-create.sh
@@ -1,6 +1,15 @@
 # Exit immediately if a command exits with a non-zero status, and print each command.
 set -e -x
 
+CLUSTER=true
+for arg in "$@"; do
+  case "$arg" in
+    --no-cluster) CLUSTER=false ;;
+  esac
+done
+
+if [ "$CLUSTER" = true ]; then
+
 echo "Setting up Minikube..."
 minikube delete
 # github actions runner has 2 cpus, 8G memory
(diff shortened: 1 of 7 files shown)


RoshDSIT:
## Code Review: PR #175 — Run "not req_k8s" tests separately

**Risk tier: Low** — CI-only change, no production code touched, fully reversible.

---

### 1. Bug — Missing `req_k8s` marker on diagnostics test (root cause of CI failure)

The no-cluster job fails on `test/diagnostics/network-issues/test_can_run_diagnostic.py::test_can_run_diagnostic_network_issues`. This test calls `run_diagnostic_eval()` which runs an actual Inspect eval with `sandbox=("k8s", "helm-values.yaml")` — it absolutely requires a k8s cluster but has no `req_k8s` marker. Since it's under `test/diagnostics/` rather than `test/k8s_sandbox/`, it was easy to miss.

**Fix:** Add `pytestmark = pytest.mark.req_k8s` to `test/diagnostics/network-issues/test_can_run_diagnostic.py` (and import pytest).

This is the only test that actually *failed* in CI, so it may be the only missing marker. However, a broader audit is warranted (see item 3 below).

---

### 2. Bug/Concern — Several `test_helm.py` tests need k8s but lack the marker

Looking at `test_helm.py`, only 2 tests are explicitly marked `@pytest.mark.req_k8s` (`test_helm_labels_misformatted` and `test_helm_labels_appear_on_release_secret`). But several other tests in that file make real helm/k8s calls without mocking:

- **`test_cancelling_install_uninstalls`** — calls `release.install()` (real helm install, no mock), then `get_all_release_names()` which calls into the k8s API
- **`test_helm_uninstall_does_not_error_for_release_not_found`** — calls `release.uninstall()` with real helm
- **`test_helm_uninstall_errors_for_other_errors`** — calls real `uninstall()` 
- **`test_helm_install_error`** — calls `release.install()` (will fail quickly because the chart doesn't exist, but still invokes real helm)
- **`test_helm_install_timeout`** — calls `release.install()` with a 1s timeout, real helm
- **`test_invalid_helm_timeout`** — calls `release.install()`, but should fail on env validation before hitting helm (probably fine)

These didn't fail in the no-cluster job because pytest runs with `-x` (stop on first failure) and the diagnostics test failed first. But if that's fixed, some of these will be the next dominoes to fall.

**The right question for the author:** Which of these tests are *intended* to work without k8s? The ones using `patch("k8s_sandbox._helm._run_subprocess", ...)` are clearly designed as unit tests. The ones calling real `install()`/`uninstall()` without mocks are not. The latter group needs `@pytest.mark.req_k8s`.

---

### 3. Concern — `test_config_validation.py` has hidden k8s dependencies

- `test_invalid_kubeconfig_context_name` calls `K8sSandboxEnvironment.sample_init()` with `context="invalid-context"` — this hits `config.list_kube_config_contexts()` which requires a kubeconfig file to exist. Without minikube, there's no kubeconfig, so this will raise `ConfigException` instead of `ValueError`.
- Several other tests in that file call `sample_init()` which may try to load k8s config depending on the code path.

Whether these need `req_k8s` depends on whether they'd survive without a kubeconfig. The test *might* work if `sample_init()` fails early on file-not-found before touching k8s config, but it's worth checking.

---

### 4. Concern — `test_prereqs.py::test_helm_version_satisfactory`

`validate_prereqs()` only checks helm version (verified in source — it's just `helm version --short`). Helm is installed by the devcontainer feature even with `--no-cluster`, so this test should pass. **No issue here.** But note that if `validate_prereqs` ever gains a k8s connectivity check, this test will silently break in the no-cluster job.

---

### 5. Nit — CI matrix: should `cluster: false` also test helm 4.1.0?

The no-cluster job only runs with helm 3.20.0. Since the "not req_k8s" tests include `test_helm.py` tests that exercise `_get_helm_major_version()` and `_get_wait_flag()` (the helm 3 vs 4 branching logic), it would be cheap to also run helm 4.1.0 without a cluster. That said, those specific tests mock `subprocess.run` anyway, so the installed helm version doesn't matter for them. **This is a judgment call** — the current split is reasonable if you want to keep CI minutes down.

---

### 6. Concern — The `sed` approach for mutating `devcontainer.json`

Two `sed` commands modify `devcontainer.json` before `devcontainer up`:

**Helm version replacement:**
```
sed -i 's/"helm": "[^"]*"/"helm": "${{ matrix.helm-version }}"/' .devcontainer/devcontainer.json
```
This works because the devcontainer.json has exactly one `"helm": "3.16.1"` occurrence. It's a bit fragile — if someone adds a comment mentioning helm versions, or if the JSON is reformatted, the regex could match incorrectly or not at all. But for a CI workflow in a small repo, this is pragmatic.

**`--no-cluster` injection:**
```
sed -i 's|bash .devcontainer/post-create.sh|bash .devcontainer/post-create.sh --no-cluster|' .devcontainer/devcontainer.json
```
This matches a literal string in `postCreateCommand`. It'll silently do nothing if that string changes (e.g., someone switches to `sh` instead of `bash`, adds flags, or moves to a `postCreateCommand` array format). **Silent failure here means the no-cluster job would spin up a full minikube cluster, wasting ~5 minutes but not breaking anything.**

**Suggestion:** Add a `grep` guard after each `sed` to verify the substitution took effect:
```yaml
- name: Configure Helm version in devcontainer.json
  run: |
    sed -i 's/"helm": "[^"]*"/"helm": "${{ matrix.helm-version }}"/' .devcontainer/devcontainer.json
    grep -q '"helm": "${{ matrix.helm-version }}"' .devcontainer/devcontainer.json
```

---

### 7. Praise — Helm version handling improvement

The old approach installed helm *after* `devcontainer up` by downloading a tarball. The new approach sets the version in devcontainer.json *before* `devcontainer up`, so the devcontainer feature installs the right version directly. This is cleaner — one fewer post-build step, and the devcontainer's helm version matches what CI actually tests. The `helm version --short` verification step is a nice sanity check.

The one difference: the devcontainer feature's helm installation mechanism might behave differently from the raw tarball download (e.g., installation path, shell completions). In practice this shouldn't matter for testing.

---

### 8. Praise — The `--no-cluster` flag

The shell script implementation is clean and robust:
- Proper `case` statement for argument parsing
- Boolean flag with string comparison (`"$CLUSTER" = true`)
- The `if`/`fi` block cleanly wraps all cluster-related setup
- `uv sync --extra dev` stays outside the conditional, which is correct

One minor observation: unknown arguments are silently ignored. That's fine for a devcontainer script.

---

### 9. Praise — `--durations=10`

Free visibility into test timing. No downsides. Good habit.

---

### 10. What Needs Human Judgment

1. **How thorough should the marker audit be before merging?** The diagnostics test is the immediate fix, but there's a spectrum from "fix the one failing test" to "audit every test and ensure perfect marker coverage." The author's PR description suggests this is partly exploratory ("They seem not to!" run without k8s), so a follow-up PR to fix markers is reasonable.

2. **Is the no-cluster job's value worth the maintenance cost?** The "not req_k8s" tests are currently: `test_limited_stream_buffer.py`, `test_logger.py`, most of `test_helm.py` (the mocked ones), `test_kubernetes_api.py` (all mocked), `test_config_validation.py` (some may need k8s), and `test_prereqs.py`. These are all fast unit tests that would run fine in the cluster jobs too. The no-cluster job's value is in *proving* they don't need k8s — a form of dependency documentation. Whether that's worth a third CI job is a team call.

3. **Should the marker default be inverted?** Right now, tests without `req_k8s` are assumed to not need k8s. An alternative is `pytestmark = pytest.mark.req_k8s` at the top of every file that needs it, with a CI lint that flags unmarked test files under `test/k8s_sandbox/` (since most tests there *do* need k8s). This would make the "oops, forgot to mark" failure mode less likely. But that's a bigger change than this PR.

---

### Summary

| # | Severity | Item |
|---|----------|------|
| 1 | **Bug** | `test_can_run_diagnostic_network_issues` needs `req_k8s` — root cause of CI failure |
| 2 | **Bug** | Several `test_helm.py` tests call real helm without `req_k8s` |
| 3 | **Concern** | `test_config_validation.py` may have hidden k8s deps |
| 4 | **OK** | `test_prereqs.py` should be fine without k8s |
| 5 | **Nit** | Consider helm 4.1.0 in no-cluster job |
| 6 | **Concern** | `sed` substitutions could silently fail — add grep guards |
| 7 | **Praise** | Helm version handling is cleaner than before |
| 8 | **Praise** | `--no-cluster` flag is well-implemented |
| 9 | **Praise** | `--durations=10` is a free win |
| 10 | **Judgment** | Marker audit scope, no-cluster job value, default-deny markers |

The PR direction is good — separating k8s-dependent from k8s-independent tests is the right call for both CI speed and test hygiene. The main blocker is fixing the missing markers so the no-cluster job actually passes.

.devcontainer/post-create.sh
```diff
@@ -1,6 +1,15 @@
...
+CLUSTER=true
+for arg in "$@"; do
+  case "$arg" in
+    --no-cluster) CLUSTER=false ;;
+  esac
+done
+
+if [ "$CLUSTER" = true ]; then
+
 echo "Setting up Minikube..."
```
bmillwood-aisi: I would advise fixing indentation in this block

.github/workflows/test.yml
```diff
@@ -9,18 +9,27 @@ jobs:
...
       fail-fast: false
       matrix:
-        helm-version: ["3.20.0", "4.1.0"]
-    name: pytest (helm ${{ matrix.helm-version }})
+        include:
+          - helm-version: "3.20.0"
+            cluster: true
+          - helm-version: "4.1.0"
+            cluster: true
+          - helm-version: "3.20.0"
```
bmillwood-aisi, line 17: maybe want to run on 4.1.0 without a cluster as well, since some of the tests use the `helm` command line even if they don't use a cluster; not strongly opinionated on this one though since I don't know anything about what the version differences that matter are

.github/workflows/test.yml
```diff
@@ -9,18 +9,27 @@ jobs:
...
+    name: pytest (helm ${{ matrix.helm-version }}${{ matrix.cluster && ', k8s' || '' }})
     steps:
       - uses: actions/checkout@v4
+      - name: Configure Helm version in devcontainer.json
+        run: |
+          sed -i 's/"helm": "[^"]*"/"helm": "${{ matrix.helm-version }}"/' .devcontainer/devcontainer.json
+      - name: Skip cluster setup
+        if: ${{ !matrix.cluster }}
+        run: |
+          sed -i 's|bash .devcontainer/post-create.sh|bash .devcontainer/post-create.sh --no-cluster|' .devcontainer/devcontainer.json
```
bmillwood-aisi, line 31: these seds are effective but I would consider something more `jq`-based, especially if we end up wanting to do anything more complex

bmillwood-aisi (commented):
just leaving some comments I had in my head at the time I stopped working on this, they are obv nonbinding

.github/workflows/test.yml
RoshDSIT: Sounds worth adding

.github/workflows/test.yml
RoshDSIT: I think we should leave as-is until it gets a notch more complex - e.g. nested/conditional JSON manipulation or something - plus it adds a dependency. If this breaks it'll fail loudly too, so no risk there.

art-dsit (approved):

PR #176, opened 2026-03-27
feat: add provider-level exec retry for transient K8s errors

rasmusfaber:
## Summary

- Adds tenacity-based retry to `K8sSandboxEnvironment.exec()` so transient K8s errors (API server rotation, dropped WebSockets) are retried transparently
- Propagates `BrokenPipeError`/`ConnectionResetError` during exec streaming as `PodError` instead of silently swallowing them
- 5 attempts with exponential jitter backoff (1-10s), matching the pattern used by Daytona and Modal providers

## Details

**Retry classification** uses positive matching on transient types with explicit exclusion of permanent types:

- **Transient (retried):** `ApiException`, `WebSocketException`, `ConnectionError`, `OSError`, `PodError`, `GetReturncodeError`
- **Permanent (not retried):** `ExecutableNotFoundError`, `RuntimeError`, `PermissionError`, `TimeoutError`

`PermissionError` and `TimeoutError` must be explicitly excluded because they are `OSError` subclasses.

The retry decorator wraps `self._pod.exec()` inside `_log_op`, so tenacity sees raw exceptions and `_log_op` only sees the final failure after retries are exhausted.

Based on the approach discussed in [inspect_ai#3468](https://github.com/UKGovernmentBEIS/inspect_ai/pull/3468) and [inspect_sandboxes@93b016d](https://github.com/meridianlabs-ai/inspect_sandboxes/commit/93b016d).

## Test plan

- [x] 14 unit tests covering transient retry, permanent no-retry, expected exception passthrough, and retry exhaustion
- [x] Existing `test_execute.py` tests updated for `PodError` propagation
- [x] Manual verification against our live K8s cluster

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,7 @@ dependencies = [
   "inspect-ai>=0.3.161",
   "kubernetes>=35.0.0",
   "jsonschema>=4.23.0",
+  "tenacity>=8.0.0",
 ]
 
 [project.optional-dependencies]
--- a/src/k8s_sandbox/_pod/execute.py
+++ b/src/k8s_sandbox/_pod/execute.py
@@ -1,5 +1,4 @@
 import base64
-import logging
 import re
 import shlex
 from contextlib import contextmanager
@@ -10,11 +9,9 @@ from inspect_ai.util import SandboxEnvironmentLimits as limits
 from kubernetes.stream.ws_client import WSClient  # type: ignore[import-untyped]
 
 from k8s_sandbox._pod.buffer import LimitedBuffer
-from k8s_sandbox._pod.error import ExecutableNotFoundError
+from k8s_sandbox._pod.error import ExecutableNotFoundError, PodError
 from k8s_sandbox._pod.get_returncode import get_returncode
 from k8s_sandbox._pod.op import PodOperation
-
-logger = logging.getLogger(__name__)
 
 COMPLETED_SENTINEL = "completed-sentinel-value"
 COMPLETED_SENTINEL_PATTERN = re.compile(rf"<{COMPLETED_SENTINEL}-(\d+)>")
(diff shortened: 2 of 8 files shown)


art-dsit:
Looks good to me. Couple of questions:

* Are the subclasses of `OSError` that are listed as permanent errors ones that you discovered through trial and error, or are they just plausible/LLM generated?
* I note the good unit test coverage; do you have a test plan already for seeing what difference this makes in a deployed cluster? (I would be happy to help design one, and it's not necessarily a blocker for merging this)

art-dsit:
My testing so far says urllib3 has its own default retry functionality which is interfering - we should probably disable this if we're using tenacity

rasmusfaber:
> * Are the subclasses of `OSError` that are listed as permanent errors ones that you discovered through trial and error, or are they just plausible/LLM generated?

They were originally imported from the code in https://github.com/UKGovernmentBEIS/inspect_ai/pull/3468 , which was built by repeatedly trying to break the runner in our infra (both manually and with red-teaming models) and fixing the resulting errors. I then reran the same tests on this code, and added the resulting errors.

> * I note the good unit test coverage; do you have a test plan already for seeing what difference this makes in a deployed cluster? (I would be happy to help design one, and it's not necessarily a blocker for merging this)

We have had a version of this deployed to our infra for some time, and it makes a large difference for long-running evals.

> My testing so far says urllib3 has its own default retry functionality which is interfering - we should probably disable this if we're using tenacity

This is for the `_check_for_pod_restart()` right? I don't think the websocket calls involve urllib3. I have hoisted that out of the tenacity loop. There is no need to re-check that on each retry anyway.

art-dsit (approved):
LGTM assuming the test failure is just flake - I'll re-run it

tadamcz:
FYI, this has made an enormous difference to our usage on MirrorCode. Long (>1bn total tokens) runs are now feasible, and were not before. Thanks!

PR #177, opened 2026-04-07
fix: prefer kubeconfig over in-cluster config to preserve configured namespace

rasmusfaber:
## Summary

- When running inside a pod with a provided kubeconfig, `load_incluster_config()` was taking priority and reading the namespace from the SA token mount (runner namespace) instead of the kubeconfig (sandbox namespace)
- This caused k8s-sandbox to install agent-env charts in the wrong namespace, where the runner SA had no RBAC permissions
- Flips the load order: try kubeconfig first, fall back to in-cluster
- Wraps both-fail case in an explicit `ConfigException` with a clear error message

## Context

This is a follow-up to #159. The original PR introduced in-cluster config detection but used in-cluster-first ordering. In METR's deployment (pods with mounted kubeconfigs), this caused the wrong namespace to be selected. This PR preserves the in-cluster support from #159 while fixing the priority order.

## Test plan

- [x] Existing tests updated to reflect kubeconfig-first ordering
- [x] New test for both-fail error case

### DIFF
--- a/src/k8s_sandbox/_kubernetes_api.py
+++ b/src/k8s_sandbox/_kubernetes_api.py
@@ -115,22 +115,31 @@ class _Config:
 
     @classmethod
     def _load(cls) -> _Config:
-        # Try in-cluster config first (running inside a pod).
+        # Try kubeconfig file first (preserves configured namespace/context).
+        try:
+            config.load_kube_config()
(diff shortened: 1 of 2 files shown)


art-dsit (approved):
LGTM assuming unit tests pass. I also did some exploratory testing

PR #179, opened 2026-04-07
feat: add configurable K8s client token refresh

rasmusfaber:
## Summary

- Adds `INSPECT_K8S_CLIENT_REFRESH_SECONDS` env var to periodically re-create cached Kubernetes API clients, causing the `kubernetes` library to re-read rotated service account tokens from disk
- Disabled by default (unset or `0`). Set to e.g. `600` to refresh clients every 10 minutes.

## Context

The `kubernetes` Python library's `KubeConfigLoader` reads the token file once and caches it in memory forever. EKS 1.35 caps JWT lifetime at 24h (down from 365 days on 1.33). Long-running evals hit 401s when the cached token expires.

The fix adds a max-age mechanism to `_ThreadLocalClientFactory`. When a cached client exceeds the configured age, it is discarded (with `api_client.close()`) and a fresh one is created via `config.new_client_from_config()`, which re-reads the token file from disk.

In-cluster mode (`load_incluster_config`) is unaffected — it already has built-in 60s token refresh.

## Test plan

Tested on METR's infra.

### DIFF
--- a/src/k8s_sandbox/_kubernetes_api.py
+++ b/src/k8s_sandbox/_kubernetes_api.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import os
 import threading
+import time
 from pathlib import Path
 from typing import TypedDict, cast
 
@@ -16,6 +17,20 @@ logger = logging.getLogger(__name__)
 _thread_local = threading.local()
 
 _INCLUSTER_NAMESPACE_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+
+INSPECT_K8S_CLIENT_REFRESH_SECONDS = "INSPECT_K8S_CLIENT_REFRESH_SECONDS"
+
+
+def _get_client_refresh_seconds() -> int:
+    name = INSPECT_K8S_CLIENT_REFRESH_SECONDS
+    raw = os.environ.get(name, "0")
+    try:
+        value = int(raw)
+    except ValueError:
+        raise ValueError(f"{name} must be a non-negative int: '{raw}'.")
+    if value < 0:
+        raise ValueError(f"{name} must be a non-negative int: '{value}'.")
+    return value
 
 
 class _KubeContext(TypedDict):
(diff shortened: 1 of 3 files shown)


art-dsit:
@rasmusfaber this has conflicts at the moment

Unfortunately I don't have a 1.35+ cluster to test this on at the moment - how critical is this change?

rasmusfaber:
> @rasmusfaber this has conflicts at the moment

Resolved.

> Unfortunately I don't have a 1.35+ cluster to test this on at the moment - how critical is this change?

It is critical enough for us that we are currently running on a hotfix branch with this feature. The actual code does not depend on K8s >= 1.35 to run. It will refresh the client with the indicated interval no matter the version.

If you are very reluctant to merge this, we do have a fallback option of switching to the incluster config instead, which automatically refreshes the token. We just felt that this was the lower risk option of the two, and it seems a natural configuration to be able to set.

src/k8s_sandbox/_kubernetes_api.py
```diff
@@ -17,6 +18,20 @@
 
 _INCLUSTER_NAMESPACE_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 
+INSPECT_K8S_CLIENT_REFRESH_SECONDS = "INSPECT_K8S_CLIENT_REFRESH_SECONDS"
+
+
+def _get_client_refresh_seconds() -> int:
+    name = INSPECT_K8S_CLIENT_REFRESH_SECONDS
```
art-dsit, line 25: I don't think having the literal string INSPECT_K8S_CLIENT_REFRESH_SECONDS three times in this file is adding much value

art-dsit:
One concern is discoverability of this config option. It seems to me the typical user's first encounter with the problem that this PR fixes will be seeing an ugly 401 in a crashed eval. Is there a way we can hook into the various places we call k8s? (But of course if that was easy, we could also fix up the token reactively instead of on a timer)

art-dsit (approved):
I can't see a better way than this PR, so happy to merge as-is to unblock you, or we can try to think of a nicer way

PR #182, opened 2026-04-08
fix: create new AsyncRetrying instance per exec call to avoid concurrency bug

rasmusfaber:
## Summary

Got #176 load tested some more, and ran into this issue when running with high concurrency. Apparently Tenacity's retry configuration are not safe for concurrency!

### Details

- `_exec_retry` was a module-level `AsyncRetrying` singleton shared across all concurrent `exec()` calls
- tenacity's `AsyncRetrying.__aiter__` [returns `self`](https://github.com/jd/tenacity/blob/main/tenacity/asyncio/__init__.py) and overwrites `_retry_state` on each iteration, so concurrent coroutines corrupt each other's retry state
- This caused `UnboundLocalError: cannot access local variable 'result'` when one coroutine's retry loop ended prematurely due to another coroutine's state being read
- Fix: change `_exec_retry` from a singleton to a factory function that returns a fresh instance per call

See [tenacity#93](https://github.com/jd/tenacity/issues/93) — the maintainer confirms async retry state is per-instance, not per-call.

### DIFF
--- a/src/k8s_sandbox/_sandbox_environment.py
+++ b/src/k8s_sandbox/_sandbox_environment.py
@@ -78,15 +78,20 @@ _PERMANENT_TYPES = (
     TimeoutError,
 )
 
-_exec_retry = AsyncRetrying(
-    stop=stop_after_attempt(5),
-    wait=wait_exponential_jitter(initial=1, max=10),
-    retry=retry_if_exception(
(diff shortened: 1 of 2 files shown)


art-dsit (approved):
LGTM, thanks!

PR #183, opened 2026-04-08
test: add two-cluster-config diagnostic for kubeconfig-first loading

art-dsit:
## Summary
- E2E diagnostic that reproduces the two-cluster scenario from PR #177: an Inspect eval running inside a pod with both in-cluster config and a mounted kubeconfig pointing at a different cluster (METR's deployment pattern)
- Two minikube clusters (runner + sandbox) verify that `_Config._load()` prefers the kubeconfig over in-cluster config
- Passes with the PR #177 fix applied, fails without it (`_Config.in_cluster = True`)

## Test plan
- `bash test/diagnostics/two-cluster-config/run.sh` — requires minikube, docker, kubectl, cilium on PATH
- Verified PASS with kubeconfig-first fix, FAIL without

### DIFF
--- a/test/diagnostics/two-cluster-config/Dockerfile
+++ b/test/diagnostics/two-cluster-config/Dockerfile
@@ -1,0 +1,11 @@
+FROM python:3.12-bookworm
+
+ARG HELM_VERSION=3.17.0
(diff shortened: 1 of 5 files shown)


art-dsit:
SLOP ALERT

PR #184, opened 2026-04-08
feat: retry transient errors in read_file and write_file

art-dsit:
## Summary
- Extends exec() retry logic (tenacity, 5 attempts, exponential jitter backoff 1-10s) to `read_file()` and `write_file()`
- Temp file is seeked/truncated between retry attempts so retries start fresh
- Adds `FileNotFoundError` and `IsADirectoryError` to permanent (non-retried) error types — both are `OSError` subclasses that would otherwise match the transient `OSError` catch
- Removes the outdated "exec not retried" section from `limitations.md`

## Test plan
- [x] 28 unit tests pass (`test/k8s_sandbox/test_exec_retry.py`) — 14 existing exec tests + 14 new read/write tests
- [x] E2e diagnostic (see #181) confirmed: FAIL without retry on parent branch, PASS with retry on this branch

### DIFF
--- a/docs/docs/design/limitations.md
+++ b/docs/docs/design/limitations.md
@@ -151,6 +151,12 @@ chart](../helm/built-in-chart.md#dns). Or, if using a custom Helm chart, conside
 the `hostAliases` field in the Pod spec
 ([docs](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/)).
 
+## Transient errors are automatically retried
+
+Transient network or infrastructure errors during `exec()`, `read_file()`, and `write_file()` are automatically retried. These are typically caused by issues such as a node becoming unhealthy, a Pod being rescheduled, or the Kubernetes control plane being overloaded.
+
+Note that retries cannot guarantee idempotency — if a command partially executed before the error, it may run again on retry.
+
 ## Must run as root to use the `user` parameter in `exec()` { #exec-user }
 
 In Kubernetes, a container runs as a single user. The user can be specified in the
--- a/docs/docs/design/limitations.md
+++ b/docs/docs/design/limitations.md
@@ -151,26 +151,6 @@ chart](../helm/built-in-chart.md#dns). Or, if using a custom Helm chart, conside
 the `hostAliases` field in the Pod spec
 ([docs](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/)).
 
-## Transient network or infrastructure issues during `exec()` won't be retried
-
-The following exceptions have occasionally been observed during calls to `exec()`,
-`read_file()` or `write_file()`:
-
-* `WebSocketBadStatusException: pod does not exist` (re-raised as `ApiException`)
-* `WebSocketBadStatusException: container not found` (re-raised as `ApiException`)
-* `WebSocketBadStatusException: Handshake status 500 Internal Server Error` (re-raised
-  as `ApiException`)
-* `WebSocketConnectionClosedException: Connection to remote host was lost`
-* `SSLEOFError: EOF occurred in violation of protocol`
-
-These are likely due to transient network or infrastructure issues. For example, when a
-node becomes unhealthy and the Pod is rescheduled.
-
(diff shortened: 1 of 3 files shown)


art-dsit:
@rasmusfaber any thoughts on this PR?

src/k8s_sandbox/_sandbox_environment.py
```diff
@@ -76,6 +76,8 @@
     RuntimeError,
     PermissionError,
     TimeoutError,
+    FileNotFoundError,
+    IsADirectoryError,
 )
 
 _exec_retry = AsyncRetrying(
```
rasmusfaber: This should probably just be `_retry`, if we use it for all operations.

rasmusfaber (approved):
Looks good. Seems clear that we should do this.

docs/docs/design/limitations.md
```diff
@@ -151,26 +151,6 @@ chart](../helm/built-in-chart.md#dns). Or, if using a custom Helm chart, conside
 the `hostAliases` field in the Pod spec
 ([docs](https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/)).
 
-## Transient network or infrastructure issues during `exec()` won't be retried
```
bmillwood-aisi, line 154: I would keep something here about the existing retry behaviour

art-dsit: changed

src/k8s_sandbox/_sandbox_environment.py
```diff
@@ -308,7 +313,11 @@ async def read_file(self, file: str, text: bool = True) -> str | bytes:
                 OutputLimitExceededError,
             )
             with self._log_op("K8s read file from Pod", expected_exceptions, file=file):
-                await self._pod.read_file(Path(file), temp_file)
+                async for attempt in _retry():
+                    with attempt:
+                        temp_file.seek(0)
+                        temp_file.truncate()
```
bmillwood-aisi, line 319: thinking this through:
- it seems like the `truncate` call here should be unnecessary, since we're going to replace any content on the next retry anyway
- ok but what if the source file changes size between retries?
- ok but that would suggest we should add truncation to `write_file` too, although in that case it happens on the remote end so I don't know if it's easy to do
- also in principle the file could change during a read anyway, assuming the file is too big to read atomically, so if the file is being concurrently edited while either of these operations are happening then getting definitely sensible things is probably hopeless

no changes requested! but thought it was useful to think through

test/k8s_sandbox/test_exec_retry.py
```diff
@@ -185,3 +208,160 @@ async def test_retries_exhausted_raises_k8s_error(self) -> None:
...
+
+
+class TestReadFileRetryExhausted:
+    """After 5 failed attempts, read_file should propagate the error."""
+
+    async def test_retries_exhausted_raises_k8s_error(self) -> None:
+        sandbox, mock_read = _make_sandbox_with_mock_read()
+        mock_read.side_effect = ApiException(status=503, reason="Service Unavailable")
+
+        with pytest.raises(K8sError):
```
bmillwood-aisi, line 293: minor: I was expecting `ApiException` here, is it wrapped somehow?

test/k8s_sandbox/test_exec_retry.py
```diff
@@ -185,3 +208,160 @@ async def test_retries_exhausted_raises_k8s_error(self) -> None:
...
+
+# -- write_file retry tests --
+
+
+class TestWriteFileRetryTransient:
+    """Transient errors during write_file should be retried."""
+
+    @pytest.mark.parametrize(
+        "error",
+        [
```
bmillwood-aisi, line 307: minor: this list is duplicated, I wonder if it's worth defining elsewhere

bmillwood-aisi (approved):

bmillwood-aisi:
as discussed on Slack, I don't like the `exec()` retrying, but this PR doesn't make me like it less, and `read_file` / `write_file` should be idempotent-ish so my objection doesn't apply to them.

PR #187, opened 2026-05-07
Don't raise runuser error when user command writes to stderr

rasmusfaber:
## Summary

Slack thread: https://inspectcommunity.slack.com/archives/C0805HJTK8U/p1778142484642579

When `exec()` is called with a `user=` parameter, the wrapper script runs the user's command via `runuser`. Afterwards, `_handle_shell_output` scans stderr for `runuser: user X does not exist` / `may not be used by non-root users` and converts those into descriptive `RuntimeError`s pointing at the docs.

This check ran on any non-zero exit, so a user-supplied command that happened to write one of those substrings to stderr, or that invoked `runuser` itself, was misreported as a configuration error rather than surfacing the command's actual failure.

The fix tracks whether the wrapper's completed-sentinel was observed on stdout. If it was, the `runuser` invocation succeeded and any stderr is from the user command, so we skip the parse. If it wasn't, the wrapper failed before the user command ran, and the existing translation is still useful.

### DIFF
--- a/src/k8s_sandbox/_pod/execute.py
+++ b/src/k8s_sandbox/_pod/execute.py
@@ -116,7 +116,7 @@ class ExecuteOperation(PodOperation):
     def _handle_shell_output(
         self, ws_client: WSClient, user: str | None, timeout: int | None
     ) -> ExecResult[str]:
-        def stream_output() -> ExecResult[str]:
+        def stream_output() -> tuple[ExecResult[str], bool]:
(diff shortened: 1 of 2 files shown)


tadamcz:
Closed mine in favour of yours! Thanks, great work.

tadamcz:
Want to open this as non-draft @rasmusfaber?

rasmusfaber:
> Want to open this as non-draft @rasmusfaber?

Ah, sorry. Forgot to move it to ready. Thanks for the heads-up.

art-dsit (approved):

PR #188, opened 2026-05-07
Use sentinel value to avoid false-positive `RuntimeError: When a user parameter is provided to exec(), must be running as root`

tadamcz:


tadamcz:
Closing in favour of Rasmus' https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/pull/187, since he has more experience with this repo

PR #189, opened 2026-05-07
test: rename bash tool argument cmd → command

art-dsit:
inspect-ai 0.3.206 renamed the `bash` tool's parameter from `cmd` to `command`.


bmillwood-aisi (approved):
merging this as-is is ok, but should we also set a minimum version in `pyproject.toml`? I guess if only the tests depend on this then maybe that's unnecessary?

another question: is there a way we could fix `tool_call` or use something different so that this is a type error instead of a runtime error?

bmillwood-aisi:
> inspect-ai 0.3.207 renamed the `bash` tool's parameter from `cmd` to `command`.

looks to me like it was [0.3.206](https://github.com/UKGovernmentBEIS/inspect_ai/blob/main/CHANGELOG.md#03206-15-april-2026) FWIW 

- see: https://github.com/UKGovernmentBEIS/inspect_ai/pull/3615

art-dsit:
> merging this as-is is ok, but should we also set a minimum version in `pyproject.toml`? I guess if only the tests depend on this then maybe that's unnecessary?

yeah I considered that but since it's only test breakage, decided to keep backward compat

> another question: is there a way we could fix `tool_call` or use something different so that this is a type error instead of a runtime error?

dunno, question for inspect core devs?

bmillwood-aisi:
> > another question: is there a way we could fix `tool_call` or use something different so that this is a type error instead of a runtime error?
> 
> dunno, question for inspect core devs?

`tool_call` is imported from `test.k8s_sandbox.inspect_integration.testing_utils.utils`, so I think it is a question for this repo. but I'm content not to spend any more time on it

PR #193, opened 2026-05-27
Recover from pod replacement instead of looping on stale UID

rasmusfaber:
## Summary

When a pod was evicted and recreated by its StatefulSet (same name, new UID), `check_for_pod_restart` raised an untyped `RuntimeError` and never refreshed the cached `PodInfo.uid`. Every subsequent operation re-read the same stale UID and re-raised, so any caller that handled or retried the error kept hitting the same dead pod for the lifetime of the sample.

This PR raises typed `PodReplacedError` / `ContainerRestartedError` (both `K8sError` subclasses) so callers can recognise the condition without string-matching, and refreshes the cached pod identity inside `Pod` when either fires. The typed exception is raised on the actual transition; subsequent operations target the new pod.

One behaviour change worth flagging: `restarted_container_behavior="warn"` now also refreshes the cached identity. Previously it logged but left the stale UID in place, which caused weird downstream "no such file" errors on the next exec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).
+
 ## 2026-05-07 0.5.0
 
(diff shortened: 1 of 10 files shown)


rasmusfaber:
The CI pre-commit error is [pre-existing](https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/actions/runs/25865434166/job/76006141415).

(Fix: https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/pull/194)

art-dsit (approved):
LGTM, thanks! I've tested it e2e and it behaves as expected

PR #190, opened 2026-05-13
Add ServiceAccount support to agent-env Helm chart

tbroadley:
## Summary
- Adds a `serviceAccountName` Helm value that, when set, creates a ServiceAccount and attaches it to all pods
- Fixes label rendering for labels containing colons
- Adds helm chart unit tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #194, opened 2026-05-27
Regenerate agent-env README (helm-docs drift)

rasmusfaber:
## Summary

`helm-docs` has drifted from `values.yaml`'s `serviceAccountName` annotation since #190 landed, which makes the `helm-docs-built` pre-commit hook fail on every subsequent PR (red CI for everyone). Pure regen: the new wording is verbatim what `helm-docs` produces from the current `values.yaml`.


art-dsit:
Thanks!

art-dsit (approved):

PR #196, opened 2026-05-28
fix: surface restart/replacement warning in the eval transcript; clarify limitations docs

art-dsit:
SLOP ALERT

## Summary

Originally a docs-only clarification of the "Containers may restart" section of `docs/docs/design/limitations.md` (in light of #193). While validating the docs end-to-end against minikube, I found the documented behaviour was wrong in one place — the `restarted_container_behavior="warn"` warning never reached the eval transcript — so this PR now also fixes that, and the docs describe the corrected behaviour.

## The bug

`Pod.check_for_pod_restart` ran detection **and** the warn/raise policy inside a `PodOpExecutor` worker thread (via `loop.run_in_executor`). In warn mode the `logger.warning(...)` was emitted from that worker. Inspect's logging-to-transcript capture is keyed on a `ContextVar` (`inspect_ai.log._transcript`), and `run_in_executor` does not propagate the context into the worker thread — so `transcript()` there resolves to a detached default and the record is dropped. Net effect: the warning showed up on stderr but **not** in the eval transcript / `inspect view`.

Isolated with a minimal `inspect eval` (separate process): the same warning logged from the main thread is captured in the transcript; logged from a `run_in_executor` worker it is not.

## The fix

Split `check_for_pod_restart` so the worker only does the blocking k8s read + identity refresh (`_detect_and_refresh`, returns the detected exception or `None`), and the async method applies the warn/raise policy on the event-loop thread. The warning is now logged from the calling thread and reaches the sample transcript. Raise behaviour is unchanged (the exception already surfaced on the event-loop thread when awaited).

## Docs changes (`limitations.md`)

- Distinguishes the two state-loss events the framework detects (pod replacement / container restart).
- Corrects what state survives each:
  - emptyDir volumes survive container restarts but not pod replacements.
  - PersistentVolume-backed paths survive both.
  - Anything on the container's writable layer is lost in both cases.
- States that the warn-mode warning is recorded in the eval transcript and is not surfaced to the agent or solver.
- Notes that in `raise` mode the sandbox `exec`/`read_file`/`write_file` methods re-wrap the typed `PodReplacedError` / `ContainerRestartedError` as a `K8sError` with the typed error as `__cause__` (catch `K8sError` and inspect `__cause__`).
- Fixes `restarted_container_behaviour` → `restarted_container_behavior`. Note the British spelling does **not** raise — the config model ignores unknown fields (pydantic `extra='ignore'`), so a copy of the old docs silently stays on the default `"warn"`.
- Corrects the `ContainerRestartedError` docstring: "Files on disk survive" was imprecise — the container writable layer is reset on restart (confirmed e2e); only mounted volumes survive (emptyDir across a restart, persistent volumes across both).

## Tests

- New unit test asserts the warn-mode warning is logged on the calling thread, not the executor worker (fails against the previous implementation, passes now).
- Existing pod-restart tests updated to drive the async `check_for_pod_restart` (raise/refresh/warn paths).
- `pytest -m "not req_k8s"` green (315 passed); ruff + mypy clean.

## E2E validation (minikube, via `inspect eval`)

Real container restart (liveness-probe failure) and real pod replacement (`kubectl delete pod`), in `warn` and `raise` modes:

- Survival matches the docs exactly — emptyDir survives restart only; PV survives both; `/root` (writable layer) lost in both.
- warn mode: refreshes identity and continues; **transcript now contains the warning** (0 LoggerEvents before the fix, 2 after).
- raise mode: typed error raised (wrapped in `K8sError`, typed error as `__cause__`); identity still refreshed so the next op succeeds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).
+- The `restarted_container_behavior="warn"` warning is now recorded in the eval transcript (visible in `inspect view`), not just on stderr. It was previously logged from a worker thread whose context Inspect's transcript capture could not see.
 
 ## 2026-05-07 0.5.0
 
--- a/docs/docs/design/limitations.md
+++ b/docs/docs/design/limitations.md
@@ -41,9 +41,27 @@ chart](../helm//built-in-chart.md#resource-requests-and-limits).
 
 You can reduce the impact of a container restarting by using persistent volumes.
 
-The framework will issue a warning if a container restarts during an eval. If you set
-the `restarted_container_behaviour` parameter to `raise`, the eval will fail the sample
-if it detects a container restart.
+The framework detects two kinds of state-loss events on every operation:
+
+* **Pod replacement** — the pod's UID has changed (the StatefulSet recreated it,
+  typically after eviction).
+* **Container restart** — the same pod's container process restarted in place
+  (`restart_count` increased).
+
+In both cases, in-memory state and any background processes started by the agent are
+lost, and anything written to the container's writable filesystem is reset. EmptyDir
+volumes survive container restarts but not pod replacements; PersistentVolume-backed
+paths survive both.
+
+By default (`restarted_container_behavior="warn"`), the framework logs a warning,
+refreshes its cached pod identity, and continues against the refreshed pod or
+container. The warning is recorded in Python `logging` and the eval transcript — it is
+not surfaced to the agent or solver.
+
+If you set `restarted_container_behavior="raise"`, the operation that detected the
+change instead raises a typed `PodReplacedError` or `ContainerRestartedError` (both
+`K8sError` subclasses). The cached identity is still refreshed, so a subsequent
+operation will target the refreshed pod.
 
 ??? question "Why not use Jobs over StatefulSets?"
 
--- a/docs/docs/design/limitations.md
+++ b/docs/docs/design/limitations.md
@@ -55,13 +55,17 @@ paths survive both.
 
 By default (`restarted_container_behavior="warn"`), the framework logs a warning,
 refreshes its cached pod identity, and continues against the refreshed pod or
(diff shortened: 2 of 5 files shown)


art-dsit:
## E2E validation (Claude)

🤖 _This validation was performed by Claude (Claude Code), driven by @art-dsit. Posting on his behalf._

I re-ran the PR's claims end-to-end against a local **minikube** cluster via `inspect eval` (real container restart triggered with `kill 1`; real pod replacement triggered with `kubectl delete pod`), in both `warn` and `raise` modes. A helm chart with three storage locations was used: the container writable layer (`/root`), an **emptyDir** mount (`/scratch`), and a **PersistentVolume** mount (`/pv`).

### Results

| Scenario | UID change | error type | `__cause__` | `/root` (writable) | `/scratch` (emptyDir) | `/pv` (PV) | warning in transcript |
|---|---|---|---|---|---|---|---|
| warn / restart  | no  | — (no raise) | — | **lost** | survived | survived | ✅ `ContainerRestartedError` |
| raise / restart | no  | `K8sError` | `ContainerRestartedError` | **lost** | survived | survived | ✅ (ERROR-level) |
| warn / replace  | yes | — (no raise) | — | **lost** | **lost** | survived | ✅ `PodReplacedError` |
| raise / replace | yes | `K8sError` | `PodReplacedError` | **lost** | **lost** | survived | ✅ (ERROR-level) |

### Confirmed
- **Survival matrix matches the docs exactly.** Writable layer reset in both events; emptyDir survives a container restart but **not** a pod replacement; PV survives both. The corrected `ContainerRestartedError` docstring ("writable layer is reset", not "files on disk survive") is right.
- **The fix works:** in `warn` mode the warning now appears as a `LoggerEvent` in the sample transcript (the bug being fixed). Verified separately that the new unit test fails against the pre-fix in-worker-thread implementation and passes after.
- **`raise` mode** re-wraps the typed error as `K8sError` with the typed error as `__cause__` (both `ContainerRestartedError` and `PodReplacedError`), and the cached identity is still refreshed, so the subsequent op succeeds.
- `pytest -m "not req_k8s"` → 315 passed; ruff + mypy clean.

### One cosmetic nit
The description says warn mode shows "2 LoggerEvents after". The count is scenario-dependent — it equals how many post-event ops run `check_for_pod_restart` before the identity refresh suppresses further warnings. My single-detecting-op repro produced **1**. The substantive claim (0 before → present after) holds; consider rewording "(0 … 2 …)" to "previously absent; now recorded" so it reproduces regardless of op count.

art-dsit:
fixed by #201

PR #198, opened 2026-05-29
Fix connection reset when piping large input to exec

rasmusfaber:
## Summary

`SandboxEnvironment.exec(input=...)` fails with `[Errno 104] Connection reset by peer` once the input reaches a few tens of MiB. The input is base64-inlined into the shell script and the entire script is sent as a single WebSocket stdin frame, which the API server rejects by resetting the connection. Inspect's checkpointing hit this injecting a ~28 MiB `restic` binary into a sandbox. This writes pod stdin in <=1 MiB frames instead.

Also cleanup the write_file path that did an unnecessarily complicated "write-temp-file", "pass IO[bytes]", "get-size-of-temp-file", "read-temp-file-back".

## Test plan

- [x] Tested end to end in a dev cluster.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).
+- Fix `exec(input=...)` failing with "Connection reset by peer" for large inputs (e.g. injecting a ~28 MiB binary): stdin is now written to the pod in ≤1 MiB WebSocket frames instead of a single oversized frame. `write_file` shares the same chunking helper.
 
 ## 2026-05-07 0.5.0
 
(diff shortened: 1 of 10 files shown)


art-dsit:
Do you want me to look at this while it's in draft or wait until marked ready for review?

Also, side note: inspect core's self_check.test_exec_input_large is meant to catch this but only does 1MB so we should bump it to 50MB or so

rasmusfaber:
> Do you want me to look at this while it's in draft or wait until marked ready for review?

Ah, sorry. Forgot to mark this as tested and move it to review.

art-dsit:
I've verified that the change to the test in https://github.com/UKGovernmentBEIS/inspect_ai/pull/4129 functions as a regression test (on main fails, on your branch passes), so maybe we could ditch the extra tests on this branch as they duplicate that?

rasmusfaber:
> I've verified that the change to the test in [UKGovernmentBEIS/inspect_ai#4129](https://github.com/UKGovernmentBEIS/inspect_ai/pull/4129) functions as a regression test (on main fails, on your branch passes), so maybe we could ditch the extra tests on this branch as they duplicate that?

I have removed `test_exec_with_large_stdin()`. Would you like me to also remove the raw unit tests?

art-dsit:
> > I've verified that the change to the test in [UKGovernmentBEIS/inspect_ai#4129](https://github.com/UKGovernmentBEIS/inspect_ai/pull/4129) functions as a regression test (on main fails, on your branch passes), so maybe we could ditch the extra tests on this branch as they duplicate that?
> 
> I have removed `test_exec_with_large_stdin()`. Would you like me to also remove the raw unit tests?

Thanks!

I wouldn't have bothered with the unit tests myself (too much mocking for my taste) but they're not doing any harm

art-dsit (approved):

PR #201, opened 2026-06-08
Propagate caller context into pod-op worker thread (fixes #200, subsumes #196 transcript fix)

### LINKED ISSUE #200: INSPECT_SANDBOX_MAX_EXEC_OUTPUT_SIZE` (and any exec-limit override) has no effect on the k8s sandbox
**Related:** inspect_ai companion fix — https://github.com/UKGovernmentBEIS/inspect_ai/pull/4141 (raises the `sandbox_service()` request-read limit; works on docker/local/daytona/modal but is neutralized on k8s by the thread hop below).

## Summary

The sandbox exec output-size limit is stored in an inspect_ai `ContextVar`. The k8s
provider reads and enforces that limit on a `ThreadPoolExecutor` worker thread that does
**not** inherit the caller's context, so any override set in the calling async context
(the `INSPECT_SANDBOX_MAX_EXEC_OUTPUT_SIZE` env var, or a programmatic
`override_max_exec_output_size()` / `_max_exec_output_size_var.set()`) is silently ignored.
The worker thread always falls back to the 10 MiB module default.

This makes the limit effectively un-tunable on k8s, and breaks features in inspect_ai that
rely on raising the cap for a specific operation (e.g. reading a large `sandbox_service()`
request payload).

## Symptom

```
INSPECT_SANDBOX_MAX_EXEC_OUTPUT_SIZE=104900000 inspect eval ...
```

The run still fails with:

```
OutputLimitExceededError: The sandbox output stream limit of 10 MiB was exceeded.
```

Note the message says **`10 MiB`**, not **`104900000 bytes`**. Since `104900000` is not a
round MiB, an effective override would have produced `"104900000 bytes"`. Seeing exactly
`10 MiB` proves the override never reached the enforcement point — it read the module
default.

## Root cause — the ContextVar does not cross the thread hop

inspect_ai stores the limit in a `ContextVar` and exposes it via a descriptor:

- `inspect_ai/util/_sandbox/limits.py`
  - `_max_exec_output_size_var: ContextVar[int]` with default
    `_DEFAULT_MAX_EXEC_OUTPUT_SIZE = 10 * 1024**2`
  - `SandboxEnvironmentLimits.MAX_EXEC_OUTPUT_SIZE` reads `_max_exec_output_size_var.get()`
    on each access
  - overrides happen in the **calling async context** via `set_sandbox_limits()` (env var)
    or a context manager that calls `.set()`

The k8s provider enforces the limit **synchronously, on a worker thread**:

- `k8s_sandbox/_pod/execute.py:120-121` —
  `stdout = LimitedBuffer(limits.MAX_EXEC_OUTPUT_SIZE)` (and `stderr`), read inside
  `_handle_shell_output.stream_output()`
- `k8s_sandbox/_pod/execute.py:206-209` —
  `raise OutputLimitExceededError(limit_str=limits.MAX_EXEC_OUTPUT_SIZE_STR, truncated_output=...)`
  when the buffer truncated
- that synchronous `stream_output()` is dispatched via
  `k8s_sandbox/_pod/executor.py` →
  `PodOpExecutor.queue_operation` →
  `await asyncio.get_event_loop().run_in_executor(self._executor, callable)`
  on a `ThreadPoolExecutor`, with **no `contextvars.copy_context()`**.

`loop.run_in_executor` does not copy the caller's context into the worker thread, and per
[PEP 567](https://peps.python.org/pep-0567/) a new thread starts with a fresh, empty
context. So the worker thread's `_max_exec_output_size_var.get()` finds no `.set()` and
returns the module default (10 MiB). The override, correctly applied in the main thread, is
never visible to the thread that reads it.

## Scope

Affects any sandbox provider that dispatches exec to a thread pool **and** reads the limit
inside that thread. Among current providers, only **k8s** does this:

- **docker / local** (inspect_ai) — read the limit in the async context (e.g.
  `inspect_ai/util/_sandbox/{local.py,docker/docker.py,exec_remote.py}`). Unaffected.
- **daytona / modal** (`inspect_sandboxes`) — `async def exec` returns full output and lets
  inspect_ai's async-context `SandboxEnvironmentProxy.verify_exec_result_size` enforce the
  limit; no worker-thread limit read. Unaffected.
- **k8s** — does its own `LimitedBuffer` + raise on the worker thread. **Affected.**

## Suggested fix

Carry the caller's context into the worker thread in `PodOpExecutor.queue_operation`
(`k8s_sandbox/_pod/executor.py`). This fixes **all** ContextVar-based limits/config, not
just this one:

```python
import contextvars

async def queue_operation(self, callable: Callable[[], T]) -> T:
    async with concurrency("pod-op", self._max_workers):
        ctx = contextvars.copy_context()
        loop = asyncio.get_event_loop()
        return await loop.run_in_executor(self._executor, lambda: ctx.run(callable))
```

`copy_context()` is evaluated inside `queue_operation`, i.e. within the live async context
that holds the override, so the worker thread runs the operation against a snapshot that
includes it.

Note: carrying the context is required precisely because inspect_ai applies these limits as
**per-operation `ContextVar.set()` overrides** in the calling async context. Approaches that
only adjust the module-level default (e.g. seeding it from the env var at import) would not
propagate a `.set()` override and so would not fix this.

## Reproduction

1. Run any eval on the k8s sandbox with
   `INSPECT_SANDBOX_MAX_EXEC_OUTPUT_SIZE=104900000` (a deliberately non-round value).
2. Have the agent produce a single command whose stdout exceeds 10 MiB.
3. Observe `OutputLimitExceededError: ... 10 MiB ...` — the override is ignored, and the
   message reports `10 MiB` rather than `104900000 bytes`.

## Context

Surfaced while fixing an inspect_ai bug where an oversized `sandbox_service()` request
payload (an agent-bridge model request carrying several multi-MB base64 images, ~12.5 MB)
overflowed the 10 MiB exec read limit. The inspect_ai-side fix
([UKGovernmentBEIS/inspect_ai#4141](https://github.com/UKGovernmentBEIS/inspect_ai/pull/4141))
raises the limit for the request read via `override_max_exec_output_size()` and, on
overflow, discards the request with an error response instead of retrying forever. That
limit raise works on docker/local/daytona/modal but is neutralized on k8s by the thread hop
described here, so on k8s the oversized request fails fast (cleanly) rather than succeeding
— this issue is what would let it succeed on k8s too.

art-dsit:
## Summary

Closes #200. PR #196 and issue #200 are two symptoms of one bug, and this fixes both at the source.

`PodOpExecutor.queue_operation` dispatched the synchronous pod operation via `run_in_executor` with no context snapshot. Per [PEP 567](https://peps.python.org/pep-0567/) a new thread starts with an empty context, so any `ContextVar` set in the calling async context was invisible to the worker thread.

inspect_ai keeps a couple of pieces of per-operation state in ContextVars that the k8s provider reads **on the worker thread**, so both were silently lost:

- **`INSPECT_SANDBOX_MAX_EXEC_OUTPUT_SIZE`** (and any `_max_exec_output_size_var` override) — the worker-thread `LimitedBuffer` (`execute.py`) read the 10 MiB module default instead of the override, making the exec output limit un-tunable on k8s (#200).
- **transcript capture** (`inspect_ai.log._transcript`) — `restarted_container_behavior="warn"` warnings, logged from the worker thread, never reached the eval transcript (#196).

## The fix

Snapshot the context inside `queue_operation` and run the operation under it:

```python
context = contextvars.copy_context()
return await asyncio.get_event_loop().run_in_executor(
    self._executor, lambda: context.run(callable)
)
```

`copy_context()` is evaluated inside `queue_operation`, within the live async context that holds the overrides, so the worker runs against a snapshot that includes them. This fixes **all** ContextVar-based config crossing the thread hop, not just these two.

## Tests

- New unit test in `test_executor.py`: a `ContextVar` set by the caller is visible inside the `queue_operation` worker callable. Fails against the previous implementation, passes now.
- `pytest -m "not req_k8s"` green (323 passed); ruff + mypy clean.

## E2E validation (minikube, via `inspect eval`, before/after)

- **Issue #200 — exec limit override:** bash command emitting 15 MB with `INSPECT_SANDBOX_MAX_EXEC_OUTPUT_SIZE=20000000`. Before: fails with `OutputLimitExceededError: ... 10 MiB ...` (worker read the default). After: succeeds (override reached the worker).
- **PR #196 — warn-mode warning in transcript:** real pod replacement (`kubectl delete pod`; the StatefulSet recreates it with a new UID), warn mode, run against this branch's `pod.py` (warning still logged from the worker thread, *not* PR #196's restructuring). Before: warning on stderr but **0** LoggerEvents in the transcript. After: **1** LoggerEvent (`Pod '...' has been replaced ...`) captured.

## Relationship to #196

This lands the transcript fix at the executor layer, which subsumes #196's logging bug. #196's `_detect_and_refresh` split is no longer required to get the warning into the transcript; its docs corrections and typed-error/raise work still stand independently. Suggest deciding how to sequence the two — e.g. rebase #196 to keep the docs + raise changes and drop (or keep, as a cleanliness change) the worker/event-loop split.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## Unreleased
 
-- Propagate the caller's context into the pod-operation worker thread (`PodOpExecutor.queue_operation` now runs the operation under a `contextvars.copy_context()` snapshot). Previously `run_in_executor` started the worker with an empty context, so ContextVar-based config set in the calling async context was silently lost. Fixes `INSPECT_SANDBOX_MAX_EXEC_OUTPUT_SIZE` (and other exec-limit overrides) having no effect on k8s, and ensures `restarted_container_behavior="warn"` warnings reach the eval transcript ([#200](https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/issues/200)).
-- Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).
+- Propagate the caller's context into the pod-operation worker thread to ensure that Inspect sandbox config overrides are honoured.
+- Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).  
 - Fix `exec(input=...)` failing with "Connection reset by peer" for large inputs (e.g. injecting a ~28 MiB binary): stdin is now written to the pod in ≤1 MiB WebSocket frames instead of a single oversized frame. `write_file` shares the same chunking helper.
 
 ## 2026-05-07 0.5.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Propagate the caller's context into the pod-operation worker thread (`PodOpExecutor.queue_operation` now runs the operation under a `contextvars.copy_context()` snapshot). Previously `run_in_executor` started the worker with an empty context, so ContextVar-based config set in the calling async context was silently lost. Fixes `INSPECT_SANDBOX_MAX_EXEC_OUTPUT_SIZE` (and other exec-limit overrides) having no effect on k8s, and ensures `restarted_container_behavior="warn"` warnings reach the eval transcript ([#200](https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/issues/200)).
 - Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).
 - Fix `exec(input=...)` failing with "Connection reset by peer" for large inputs (e.g. injecting a ~28 MiB binary): stdin is now written to the pod in ≤1 MiB WebSocket frames instead of a single oversized frame. `write_file` shares the same chunking helper.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Propagate the caller's context into the pod-operation worker thread to ensure that Inspect sandbox config overrides are honoured.
-- Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).  
+- Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Propagate the caller's context into the pod-operation worker thread (`PodOpExecutor.queue_operation` now runs the operation under a `contextvars.copy_context()` snapshot). Previously `run_in_executor` started the worker with an empty context, so ContextVar-based config set in the calling async context was silently lost. Fixes `INSPECT_SANDBOX_MAX_EXEC_OUTPUT_SIZE` (and other exec-limit overrides) having no effect on k8s, and ensures `restarted_container_behavior="warn"` warnings reach the eval transcript ([#200](https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/issues/200)).
```
art-dsit: this is ridiculously verbose, we need tighter changelog entries

dpolatajko-aisi (approved):
lgtm !

PR #202, opened 2026-06-11
feat: surface container crash diagnostics on Helm install failure

dmitrii:
## Problem

When a sandbox pod fails to become ready, the error surfaced to the caller contains
only Helm's generic symptom, which often does not distinguish between different
failures or doesn't convey the details that it could. Observed on a local `kind` 
cluster (helm v4.2.0):

| Failure | What Helm reports |
| --- | --- |
| Bad image | `Helm install timed out (context deadline exceeded) … Pod is in the Pending phase` |
| Unschedulable | `Pod could not be scheduled` |
| App crash (exit 1) | `Containers in CrashLoop state: default` |
| OOM kill (exit 137) | `Containers in CrashLoop state: default` |

The actionable details (`ImagePullBackOff`, `OOMKilled`, exit code, the failing image, 
the `FailedScheduling` reason) are present in the Kubernetes API but is not propagated.

## Change

On Helm install failure, read the release's pod container states and recent `Warning`
events from the Kubernetes API and append them to the raised error as `pod_diagnostics`
(alongside the existing `release` and `result`). After this change, the same four
failures now include:

| Failure | Added `pod_diagnostics` |
| --- | --- |
| Bad image | `waiting (ImagePullBackOff: … ErrImagePull: …) [image: nonexistent.invalid/no-such-image:latest]` |
| Unschedulable | `event (FailedScheduling): 0/1 nodes are available: 1 Insufficient cpu` |
| App crash | `last terminated Error (exit code 1), restarted N time(s)` |
| OOM kill | `terminated OOMKilled (exit code 137), restarted N time(s)` |

## Safety

- Diagnostics gathering is **best-effort**: the whole helper is wrapped so any failure
  while collecting them returns `None` and the original install error is raised
  unchanged. It can never mask the real failure.
- **Container/pod states and event messages only — no container log content** is read,
  to avoid surfacing agent-generated or sensitive output.

## Async / blocking IO

The Kubernetes Python client is synchronous. Rather than introduce an async-client
dependency, the diagnostics helper stays synchronous and is run in a thread via
`run_in_executor` at the call site, matching how `get_sandbox_pods` and
`_watch_for_scheduling_events` already offload blocking K8s calls. `_raise_install_error`
becomes `async` accordingly.

Alternatives considered:
- **Per-call `run_in_executor` inside the helper** (mirroring `get_sandbox_pods`
  line-for-line) — viable, but keeping `_diagnostics.py` free of `asyncio` keeps it pure
  and unit-testable without an event loop (the diagnostics tests are plain synchronous
  functions), and wraps the two K8s calls in a single thread hop.
  - **Async client (`kubernetes_asyncio`)** — this would be a new dependency and a second
  auth/config path, inconsistent with the rest of the package, it seems, which uses the
  synchronous client throughout.

## Testing

- Unit tests (`test/k8s_sandbox/test_diagnostics.py`) cover OOMKilled, ImagePullBackOff,
  FailedScheduling events (including the server-side `field_selector="type=Warning"`),
  the error-safety guarantee, and the healthy/no-op case, using real
  `kubernetes.client` model objects with only the API boundary mocked.
- Integration tests (`test/k8s_sandbox/test_helm.py`) confirm the diagnostics are
  threaded into the raised error and omitted cleanly when unavailable.
- Verified end-to-end on a `kind` cluster (helm v4.2.0): drove the real
  `Release.install()` path against deliberately-failing pods (image-pull, crash-loop,
  unschedulable, OOM) and confirmed the `pod_diagnostics` above, including that
  `field_selector="type=Warning"` returns the `FailedScheduling`/`Failed`/`BackOff`
  events.
- `ruff check`, `ruff format`, `mypy`, and the full unit suite all pass.

### DIFF
--- a/src/k8s_sandbox/_diagnostics.py
+++ b/src/k8s_sandbox/_diagnostics.py
@@ -1,0 +1,117 @@
+from __future__ import annotations
+
+import logging
+
+from kubernetes.client import CoreV1Api, V1ContainerStatus  # type: ignore
+
+from k8s_sandbox._kubernetes_api import k8s_client
+
+logger = logging.getLogger(__name__)
+
+
+def describe_release_pods(
+    context_name: str | None, namespace: str, release_name: str
+) -> str | None:
+    """Summarise the state of a Helm release's pods for inclusion in error messages.
+
+    Reads container states, termination reasons and exit codes from the Kubernetes API
+    so that failures which Helm reports only generically (e.g. "Pod is in the Pending
+    phase" or "Containers in CrashLoop state") can be attributed to a concrete cause
+    such as ImagePullBackOff or OOMKilled.
+
+    This is best-effort: it is intended to be called from error-handling paths, so any
+    failure while gathering diagnostics is swallowed and results in None rather than
+    masking the original error.
+
+    Args:
+        context_name: The kubeconfig context name, or None for the current context.
+        namespace: The namespace the release was installed into.
+        release_name: The Helm release name (used to select its pods).
+
+    Returns:
+        A human-readable, multi-line summary, or None if no useful diagnostics could be
+        gathered.
+    """
+    try:
+        return _collect_diagnostics(context_name, namespace, release_name)
+    except Exception:
+        logger.debug("Failed to collect pod diagnostics.", exc_info=True)
+        return None
+
+
+def _collect_diagnostics(
+    context_name: str | None, namespace: str, release_name: str
+) -> str | None:
+    client = k8s_client(context_name)
+    pods = client.list_namespaced_pod(
+        namespace, label_selector=f"app.kubernetes.io/instance={release_name}"
+    )
+    lines: list[str] = []
+    pod_names: set[str] = set()
+    for pod in pods.items:
+        if pod.metadata is None:
+            continue
+        pod_names.add(pod.metadata.name)
+        if pod.status is None:
+            continue
+        for container in pod.status.container_statuses or []:
+            line = _describe_container(container)
+            if line is not None:
+                lines.append(line)
+
+    lines.extend(_describe_warning_events(client, namespace, pod_names))
+
+    if not lines:
+        return None
+    return "\n".join(lines)
+
+
+def _describe_warning_events(
+    client: CoreV1Api, namespace: str, pod_names: set[str]
+) -> list[str]:
+    """Return formatted Warning events that involve any of the given pods."""
+    events = client.list_namespaced_event(namespace, field_selector="type=Warning")
+    lines: list[str] = []
+    for event in events.items:
+        involved = event.involved_object
+        if involved is None or involved.name not in pod_names:
+            continue
+        lines.append(f"event ({event.reason}): {event.message}")
+    return lines
+
+
+def _describe_container(container: V1ContainerStatus) -> str | None:
+    """Describe a single container's problematic state, or None if it looks healthy."""
+    state = container.state
+    last_state = container.last_state
+    waiting = state.waiting if state else None
+    terminated = state.terminated if state else None
+    last_terminated = last_state.terminated if last_state else None
+
+    parts: list[str] = []
+    if waiting is not None:
+        detail = waiting.reason or "Waiting"
+        if waiting.message:
+            detail += f": {waiting.message}"
+        parts.append(f"waiting ({detail})")
+    if terminated is not None:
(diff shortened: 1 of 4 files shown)


RoshDSIT:
Thanks, we'll review this week.

src/k8s_sandbox/_diagnostics.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+    lines: list[str] = []
+    pod_names: set[str] = set()
+    for pod in pods.items:
+        if pod.metadata is None:
+            continue
+        pod_names.add(pod.metadata.name)
+        if pod.status is None:
+            continue
+        for container in pod.status.container_statuses or []:
```
RoshDSIT, line 64: Could you add init container statuses to this as well please @dmitrii?
On forcing a broken init container in local testing, `describe_release_pods` only shows `'container 'default': waiting (PodInitializing)`.

RoshDSIT (changes requested):
Tested locally on minikube and the image-pull, OOM and crash-loop diagnostics surface the right info nicely.

One gap I noticed on local testing  - `_collect_diagnostics` only reads `container_statuses` so a failing init container shows as waiting (`PodInitializing`) rather than the actual cause. It'd be good to cover `init_container_statuses` too imo.

@art-dsit could you take a look when you have a moment? I've tested the cases above and left one comment. Thanks.

art-dsit (approved):
LGTM. I agree with @RoshDSIT 's comment about init containers, would be good to fix before merge

dmitrii:
Thank you for reviewing, @RoshDSIT and @art-dsit, I'll make the update.

dmitrii:
Done! ` _collect_diagnostics` now reads `init_container_statuses` too, with those lines labelled `init container '<name>'`.

Verified on `kind` against a pod with a deliberately-failing init container (exit 1):

Before:
```
container 'default': waiting (PodInitializing)
```

After:
```
init container 'broken-setup': terminated Error (exit code 1), restarted 3 time(s) [image: ...]
container 'default': waiting (PodInitializing)
```

Added a unit test for the init-container path. Anything else?

src/k8s_sandbox/_diagnostics.py
```diff
@@ -0,0 +1,127 @@
...
+    context_name: str | None, namespace: str, release_name: str
+) -> str | None:
+    client = k8s_client(context_name)
+    pods = client.list_namespaced_pod(
+        namespace, label_selector=f"app.kubernetes.io/instance={release_name}"
+    )
+    lines: list[str] = []
+    pod_names: set[str] = set()
+    for pod in pods.items:
+        if pod.metadata is None:
```
RoshDSIT: mypy: `pod.metadata.name` is `str | None` going into `pod_names.add()`. A pod with no name can't be matched against events anyway, so it's safe to skip here.

```suggestion
        if pod.metadata is None or pod.metadata.name is None:
```

test/k8s_sandbox/test_diagnostics.py
```diff
@@ -0,0 +1,189 @@
+from unittest.mock import MagicMock, patch
+
+from kubernetes.client import (  # type: ignore
+    CoreV1Event,
+    CoreV1EventList,
+    V1ContainerState,
```
RoshDSIT, line 6: Companion import for the suggestion below.

```suggestion
    V1ContainerState,
    V1ContainerStateRunning,
```

test/k8s_sandbox/test_diagnostics.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+def test_returns_none_when_all_containers_healthy_and_no_events() -> None:
+    container = V1ContainerStatus(
+        name="default",
+        image="busybox:1.36",
+        image_id="",
+        ready=True,
+        restart_count=0,
+        state=V1ContainerState(running={"started_at": None}),
```
RoshDSIT: mypy wants the typed model here rather than a raw dict, which also matches how the other tests in this file construct states.

```suggestion
        state=V1ContainerState(running=V1ContainerStateRunning(started_at=None)),
```

RoshDSIT (commented):
CI has now run on this PR (fork workflow runs needed approval on our side, which I'd missed), and mypy caught two small things which I have committed on top of your changes. Thanks again for the init container change.

RoshDSIT (approved):

PR #204, opened 2026-06-16
fix: recheck for container restart after failed exec

### LINKED ISSUE #191: Container restart during streamed `exec_remote` surfaces as opaque `RetryError`
I'm running a long-running `inspect_swe.claude_code()` agent inside a k8 sandbox. When the container gets OOM-killed and the StatefulSet restarts it in place, the next poll fails with:

```
RuntimeError: Sandbox.exec failure executing exec_remote_poll(pid: 127, ack_seq: 1032):
timeout: failed to run command '/var/tmp/.da7be258e003d428/inspect-sandbox-tools':
No such file or directory
```

followed (after 5 retries / 30s) by `tenacity.RetryError`.

The sample failing due to OOM is inevitable, but the logging and error raised would ideally be clearer to aid in diagnostics.

Claude's understanding of the causes: The k8 package has support for detecting restarts (`check_for_pod_restart`) for some entry points (`_pod/read.py:20`, `_pod/write.py:19`, `_sandbox_environment.py:271`). The streaming-poll loop in `inspect_ai.util._sandbox.exec_remote._poll` RPCs directly to `inspect-sandbox-tools` via `SandboxJSONRPCTransport` — it doesn't go through those entry points.

RoshDSIT:
Closes #191

Adds a post-exec check in `Pod.exec()` for container restarts when exec fails.

If `exec()` fails, and the pre-check found that the container had restarted, we re-raise the initial error.
If `exec()` doesn't initially fail, but exec fails, and the post-check finds the container has restarted, we raise `ContainerRestartedError`/`PodReplacedError` that was detected.

API errors during the post-check are logged at WARNING and don't mask the failure.

### DIFF
--- a/src/k8s_sandbox/_pod/pod.py
+++ b/src/k8s_sandbox/_pod/pod.py
@@ -161,8 +161,7 @@ class Pod:
             raise
         except Exception:
             logger.warning(
-                "Post-exec restart re-check failed; "
-                "returning original exec result",
+                "Post-exec restart re-check failed; returning original exec result",
                 exc_info=True,
             )
 
--- a/src/k8s_sandbox/_pod/pod.py
+++ b/src/k8s_sandbox/_pod/pod.py
@@ -151,13 +151,12 @@ class Pod:
         if not result.success:
             if warned_restart is not None:
                 raise warned_restart
-            await self._recheck_for_restart_after_failed_exec()
+            await self._diagnose_restart_after_failed_exec()
         return result
 
-    async def _recheck_for_restart_after_failed_exec(self) -> None:
-        """Re-check for restart after a failed exec to close the race window."""
+    async def _diagnose_restart_after_failed_exec(self) -> None:
         try:
-            await self._run_async(self._recheck_for_restart_sync)
+            restart = await self.check_for_pod_restart()
         except (PodReplacedError, ContainerRestartedError):
             raise
         except Exception:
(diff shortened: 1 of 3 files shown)


emilschmitz:
I don't know if that's intended, but I think this doesn't raise if the API already updated with the restart count before the first check. See e.g. this
```
async def test_warn_mode_restart_before_exec_does_not_raise():
    pod = _make_pod(restarted_container_behavior="warn")
    k8s_pod = _k8s_pod(uid="uid-OLD", container_name="default", restart_count=1)

    with (
        patch("k8s_sandbox._pod.op.k8s_client") as mock_client,
        patch("k8s_sandbox._pod.pod.ExecuteOperation") as mock_exec_cls,
    ):
        mock_client.return_value.read_namespaced_pod.return_value = k8s_pod
        mock_exec_cls.return_value.exec.return_value = _failed_exec_result()

        with pytest.raises(ContainerRestartedError):
            await pod.exec(
                cmd=["echo", "hello"],
                stdin=None,
                cwd=None,
                env={},
                user=None,
                timeout=None,
            )

```
which fails.

src/k8s_sandbox/_pod/pod.py
```diff
@@ -140,13 +143,31 @@ async def exec(
...
+            if warned_restart is not None:
+                raise warned_restart
+            await self._diagnose_restart_after_failed_exec()
         return result
 
+    async def _diagnose_restart_after_failed_exec(self) -> None:
+        try:
+            restart = await self.check_for_pod_restart()
+        except (PodReplacedError, ContainerRestartedError):
+            raise
```
dpolatajko-aisi, line 161: from `docs/design/limitations.md`:

> The framework will issue a warning if a container restarts during an eval. If you set
the `restarted_container_behaviour` parameter to `raise`, the eval will fail the sample
if it detects a container restart.

do i understand correctly that is no longer correct after this change? if so, we should document accordingly

RoshDSIT: Thanks - I'll update to document the `restarted_container_behaviour="warn"` behaviour

dpolatajko-aisi (approved):

PR #207, opened 2026-06-18
Add per-service `x-inspect_k8s_sandbox.resources` extension to the compose converter

jjallaire:
## Summary

Adds a per-service `x-inspect_k8s_sandbox` extension to the compose-to-helm
converter so a Docker Compose service can declare Kubernetes resource
`requests`/`limits` that the existing Compose shortcuts cannot express — most
notably a **request-only** resource such as `ephemeral-storage` (a scheduling
floor with no eviction cap).

```yaml
services:
  default:
    image: example.com/foo@sha256:...
    mem_limit: 32g
    cpus: 4.0
    network_mode: none
    x-inspect_k8s_sandbox:
      resources:
        requests:
          ephemeral-storage: 80Gi
```

converts to:

```yaml
services:
  default:
    resources:
      limits:   {memory: 32Gi, cpu: 4.0}
      requests: {memory: 32Gi, cpu: 4.0, ephemeral-storage: 80Gi}
```

## Motivation

`ephemeral-storage` is a first-class Kubernetes resource, and a request with no
matching limit ("schedule me where there's ≥80Gi free disk, but don't cap/evict
me") is a legitimate pattern. None of the converter's existing surfaces can
express it:

- A bare service-level `x-ephemeral-storage` is rejected by the converter's
  leftover-key guard.
- `deploy.resources.reservations.ephemeral-storage` fails compose-spec schema
  validation — `deploy.resources.{limits,reservations}` is a closed list
  (`cpus`/`memory`/`pids`/`generic_resources`/`devices`,
  `additionalProperties: false`).
- The root `x-inspect_k8s_sandbox` handler is a closed allowlist
  (`allow_domains`/`allow_entities`), and services had no extension handler at
  all.

Docker Compose has no equivalent to map from: it covers CPU and memory
requests/limits but has no disk (`ephemeral-storage`) request/reservation concept,
and the closest analog — `storage_opt.size` — is a storage-driver-dependent limit
on the container's writable layer, which is neither a request nor the same scope as
Kubernetes ephemeral storage. The extension is therefore the supported surface for
these resources.

## What changed

- **`src/k8s_sandbox/compose/_converter.py`** — `_ServiceConverter._convert_service`
  now consumes a per-service `x-inspect_k8s_sandbox` extension (after the
  `resources` block is built from `mem_limit`/`cpus`/`deploy`, before the
  leftover-key guard). Two new helpers, `_apply_service_extensions` and
  `_merge_extension_resources`, deep-merge the extension's `resources` block into
  the existing `requests`/`limits`. Resource names and values are passed through
  generically (not hardcoded to `ephemeral-storage`), so `hugepages-2Mi` etc. work
  too.

- **`test/k8s_sandbox/compose/test_converter.py`** — 7 new tests: request-only
  (no mem/cpu), merge alongside `mem_limit`/`cpus`, arbitrary resource names,
  unsupported extension key, unsupported `resources` sub-key, conflict detection,
  and a round-trip validation against the chart's `values.schema.json`.

- **`docs/docs/helm/compose-to-helm.md`** — new "Resource Requests and Limits"
  section documenting the extension and why Compose has no equivalent.

- **`CHANGELOG.md`** — Unreleased entry.

No schema changes are required: the compose-spec `service` definition already
allows `^x-` keys, and the chart's `values.schema.json` types
`services.*.resources` as an opaque object (rendered via `toYaml`), so the merged
keys validate and template unchanged.

## Design decisions

- **Raise on conflict.** If the extension sets a `requests`/`limits` key already
  populated by a Compose shortcut (e.g. `requests.memory` alongside `mem_limit`),
  the converter raises `ComposeConverterError` rather than silently overriding —
  consistent with the converter's strict reject-on-ambiguity style.
- **Scoped, not flat or wide.** The surface mirrors the Kubernetes `resources`
  structure exactly (self-documenting, generalizes to any resource) but stays
  bounded to `resources` — it is not a flat single-purpose key (which would hide
  the request-vs-limit semantics) nor an arbitrary Helm-values pass-through (which
  would duplicate fields the converter already derives and erode the
  "compose file vs. Helm values file" boundary).
- **No request→limit mirroring for extension keys.** The existing
  `_set_requests_to_limits_if_unset` only copies limits→requests, so a
  request-only `ephemeral-storage` is intentionally not mirrored into `limits`.

## Testing

- `pytest test/k8s_sandbox/compose/ -m "not req_k8s"` — 102 passed (incl. 7 new).
- `mypy` and `ruff check`/`ruff format` clean on the changed files.
- End-to-end conversion snippet and `values.schema.json` round-trip both verified.

No Kubernetes cluster is required; all converter tests are non-`req_k8s`.

## Out of scope

Downstream consumers (e.g. setting the prefixed `x-inspect_k8s_sandbox` key on an
inspect_ai `ComposeService` to drop manual values-file patching) are separate
follow-ups in their own repos.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add a per-service `x-inspect_k8s_sandbox.resources` extension to the compose-to-helm converter for declaring Kubernetes resource `requests`/`limits` (e.g. request-only `ephemeral-storage`) that the `mem_limit`/`cpus`/`deploy.resources` shortcuts cannot express. Keys are merged into the resources derived from those shortcuts; conflicts are rejected rather than silently overridden.
 - Propagate the caller's context into the pod-operation worker thread to ensure that Inspect sandbox config overrides are honoured.
 - Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).
 - Fix `exec(input=...)` failing with "Connection reset by peer" for large inputs (e.g. injecting a ~28 MiB binary): stdin is now written to the pod in ≤1 MiB WebSocket frames instead of a single oversized frame. `write_file` shares the same chunking helper.
--- a/docs/docs/helm/compose-to-helm.md
+++ b/docs/docs/helm/compose-to-helm.md
@@ -102,3 +102,45 @@ services:
     image: ubuntu
     network_mode: none
 ```
+
+## Resource Requests and Limits
+
+Use the per-service `x-inspect_k8s_sandbox` extension to set Kubernetes resource
+requests and limits that the Docker Compose shortcuts (`mem_limit`, `cpus` and
+`deploy.resources`) cannot express. The `resources` block is merged into the
+`requests`/`limits` that the converter derives from those shortcuts.
+
+The motivating case is a _request-only_ resource such as `ephemeral-storage` — a
+scheduling floor with no eviction cap:
+
+```yaml
+services:
+  default:
+    image: ubuntu
+    mem_limit: 32g
+    cpus: 4.0
+    x-inspect_k8s_sandbox:
+      resources:
+        requests:
+          ephemeral-storage: 80Gi
+```
+
+This produces the following Helm values for the service:
+
+```yaml
+services:
+  default:
+    resources:
+      limits: {memory: 32Gi, cpu: 4.0}
+      requests: {memory: 32Gi, cpu: 4.0, ephemeral-storage: 80Gi}
+```
+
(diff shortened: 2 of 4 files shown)


RoshDSIT:
A few notes:
- The `x-k8s` alias doesn't appear in the PR body or in `compose-to-helm.md`. Worth documenting as supported option.
- Noting that the passthrough means values aren't validated at conversion time, which feels like the right decision over reimplementing that here.
- `x-inspect_k8s_sandbox` now means different things at different levels - `allow_{domains, entities}` at root level, and `resources` at service level, but the tests do document it. Might be worth adding a line in the docs however.

jjallaire:
Yes, x-k8s shorthand is purposely left out of the docs (for now) as it will be a head fake for those running older versions. We should make this change in ~ a month.

As noted in the review, payload in service is purposely invalidated since it's a pass through to the full helm config.

x-k8s across multiple contexts with different payload is fine. The semantic is for a generic provider pass through not a specific schema.

art-dsit (approved):

PR #208, opened 2026-06-19
Harden allowDomains to block domain fronting

seddy-aisi:
**BREAKING CHANGE** There are two important breaking changes to be aware of:

1. `allowDomains` configurations that use ports other than 80/443 connections will need additional `allowDomainsPorts` configuration (e.g. ssh cloning from github)
2. Domains using wildcards in `allowDomains` will break on cilium < 1.18.1

This change restricts network egress by enforcing [SNI](https://docs.cilium.io/en/latest/security/policy/layer4/#limit-tls-server-name-indication-sni) (for HTTPS) and [Host](https://docs.cilium.io/en/latest/security/policy/layer7/#http) (for HTTP) checks in cilium. This prevents a possible vector where [domain fronting](https://www.zscaler.com/blogs/security-research/analysis-domain-fronting-technique-abuse-and-hiding-cdns) is possible by curation of HTTP requests.

Any existing`allowDomains` configuration will still work for standard HTTP(S) calls, which is likely its main use (to allow evals to install libraries or access websearches). I've added an additional `allowdomainsPorts` configuration for special cases where people may have been relying on other ports. For example, port 22 might be desirable for ssh git cloning or allowing evals to access external bastions.

Note that cilium <1.18 doesn't support wildcard matching on SNI, so we recommend upgrading cilium if this is required.

Domain fronting could allow relatively unrestricted internet access which is a significant security flaw, so I think these breaking changes are reasonable.

Part of RESE-690


docs/docs/helm/built-in-chart.md
```diff
@@ -111,8 +111,18 @@ from a variety of sources:
       - "security.ubuntu.com"
       - "mirror.vinehost.net"
       - "*.rubygems.org"
+    allowDomainsPorts:
```
art-dsit: would it be horrific, if instead of a separate 'allowDomainsPorts', we extended the syntax of allowDomains to add an optional port? Or does this complicate the templating later on too much?

'pypi.org'
'github.com:22'

I think we can assume it's always TCP

src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
```diff
@@ -46,9 +46,69 @@ spec:
...
+    {{- end }}
     - toFQDNs:
     {{- range .Values.allowDomains }}
       - matchPattern: "{{ . }}"
+    {{- end }}
+      {{- if has "*" .Values.allowDomains }}
+      # allowDomains includes "*" (allow all): identity enforcement is moot and
+      # serverNames has no allow-all form, so all resolved IPs are permitted.
+      {{- else }}
+      # Enforce the request identity on the pinned IPs (TLS SNI on 443, HTTP Host
```
art-dsit, line 67: I'm concerned that [ECH](https://support.mozilla.org/en-US/kb/understand-encrypted-client-hello) won't be compatible with this SNI inspection? Are we just out of luck in that case?

art-dsit (commented):
I'm not entirely convinced we should fix this, given the fix seems kinda brittle to me with the ECH problem. 

It might be better simply to note that allowDomains has this leak. If you're an eval author you can check whether the domain in question is hosted by a CDN such as Cloudfront / Cloudflare / etc. and decide whether you want to take that risk.

Or, since we were planning another piece of work to intercept / MITM all TLS streams anyway, which might be more robust, could we wait until then?

docs/docs/helm/built-in-chart.md
seddy-aisi: I'm just trying to leave things as configurable as possible given the current state of the world is "full access on all ports on these domains", because the status quo is just restricting at a DNS level. Consequently, allowing people to use UDP  if they want (e.g. for QUIC on HTTP/3 which currently works or allow specific telemetry) seems like a sensible level of flexibility to leave in place.

I personally prefer the explicitness of this approach to the `github.com:22` shorthand, but that's arguably just personal preference, happy to revisit if you have a strong view on it

src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
seddy-aisi: I seem to recall looking at this and thinking ECH wouldn't be blocked but I may be misremembering or confusing with something else - I can see if I can reproduce this though, it's worth knowing how it behaves

docs/docs/helm/built-in-chart.md
art-dsit: hmm yeah QUIC could also be a problem https://github.com/cilium/cilium/issues/28497

src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
art-dsit: ECH has an outer SNI, which *can* be read by cilium, but this is usually set to a generic domain like cloudflare-ech.com, so as long eval authors don't add that to allowDomains, it's OK.

art-dsit: I assume that as it stands, we return NXDOMAIN for all non-allowDomains, so the instructions here
https://developers.cloudflare.com/ssl/edge-certificates/ech/#enterprise-network-applicability are implicitly already followed. This deals with the case of a well-behaved client; adversarial clients will not be able to get past cilium since it will fail closed

src/k8s_sandbox/resources/helm/agent-env/README.md
```diff
@@ -8,7 +8,8 @@
...
-| allowDomains | list | Empty list (no internet access) | A list of fully qualified domain names that pods within the agent environment are allowed to access. |
+| allowDomains | list | Empty list (no internet access) | Domains the agent environment may reach. Egress is restricted to ports 80/443 and the request identity must match an entry (TLS SNI on 443, HTTP Host on 80). Use allowDomainsPorts to open other ports. Wildcard entries require Cilium >= 1.18. |
+| allowDomainsPorts | list | Empty list (only 80/443 are reachable on allowedDomains) | Extra ports opened to the resolved IPs of allowDomains, as a list of {port, protocol, domain}. protocol defaults to ANY (TCP+UDP); domain (optional) scopes the port to one allowDomains entry, otherwise it applies to all of them. These are pinned by IP only with no SNI/Host check, so only open ports you trust on those IPs. 80 and 443 may only be listed with protocol UDP (e.g. QUIC); TCP/ANY there is rejected as it would disable the identity check above. |
```
art-dsit: I think the wording here is a little too dense; maybe this deserves a separate paragraph outside the table. I don't think it's especially obvious how adding TCP/ANY on 443 would disable the identity check

docs/docs/helm/built-in-chart.md
tcatling: i'm not sure if this will work as you expect because allowDomains relies on a DNS lookup to add the IP to the list, and I think is only going to apply to http traffic because it's an L7 filter. Does this

```
# Optionally allow specific ports, example to allow ssh cloning from github.com
      - port: 22
        protocol: TCP
        domain: github.com # NOTE: must be in allowDomains list
```

actually work reliably?

tcatling: specifically:
* does it work if you _only_ include the host in allowedDomains, or does it also rely on the client (or someone else on the same host...) having done a DNS lookup on that domain
* does it actually work at all, because i think we're mixing L7 HTTP rules with SSH

seddy-aisi: @tcatling - not sure I follow? As per the comment, github.com has to be in `allowDomains` for `allowDomainsPorts` to be used

tcatling: sorry i think it's fine. I was worried if, for example, SSH to github.com might requires a separate HTTP request to github.com first to get the IPs cached, but ofc the SSH client is doing it's own DNS lookup so it should be fine.

docs/docs/helm/built-in-chart.md
```diff
@@ -98,7 +98,7 @@ from a variety of sources:
       default:
         image: ubuntu:24.04
         command: ["tail", "-f", "/dev/null"]
-    allowDomains:
+    allowDomains: # TCP on ports 80/443 allowed on all domains
```
tcatling: it actually becomes HTTP(S) only at this point right, since we rely on SNI from TLS handshake

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- **BREAKING CHANGE**: `allowDomains` egress is now restricted to ports 80/443 with the request identity enforced in Cilium (TLS SNI on 443, HTTP `Host` on 80), not just the resolved IP. Wildcard entries require Cilium >= 1.18 (older Cilium matches exact hostnames only). New `allowDomainsPorts` opens additional ports to those domains (IP-pinned only; see `values.yaml`).
```
tcatling: I think worth being explicit that allowDomains now refers to HTTP(S) traffic **only**. This mentions TLS but I think some readers will need us to be more specific.

tcatling:
I'll defer to @art-dsit and co for approval on this but it seems reasonable to me.

I think it's worth being explicit that `allowDomains` --> for HTTP(S) traffic, `allowDomainsPorts` --> for allowing arbitrary TCP traffic to those domains. The naming is somewhat confusing here but I don't see a better options which retains backwards compatibility.

src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
seddy-aisi: Yup, exactly

docs/docs/helm/built-in-chart.md
seddy-aisi: Nope, HTTP is still allowed

art-dsit (approved):

docs/docs/helm/built-in-chart.md
seddy-aisi: And it definitely works; this is required for `apt install` to work, so I explicitly tested

PR #209, opened 2026-06-22
fix: replace non-utf-8 outputs in exec

dpolatajko-aisi:
This PR updates the `exec` function to handle binary characters with replacement to their utf-8 equivalent, rather than raising. This is consistent with the current behaviour of local and docker sandboxes, and reflects the update documented in [this Slack thread](https://inspectcommunity.slack.com/archives/C0855KJ6BMW/p1782122706621579)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Fix `exec()` raising `UnicodeDecodeError` when a command writes non-UTF-8 bytes to stdout/stderr. Invalid bytes are now replaced with the unicode replacement character (matching Inspect's other sandboxes) instead of aborting the call.
+- Support replacement rather than throwing `UnicodeDecodeError` when a command writes non-UTF-8 bytes to stdout/stderr.
 - Propagate the caller's context into the pod-operation worker thread to ensure that Inspect sandbox config overrides are honoured.
 - Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).
 - Fix `exec(input=...)` failing with "Connection reset by peer" for large inputs (e.g. injecting a ~28 MiB binary): stdin is now written to the pod in ≤1 MiB WebSocket frames instead of a single oversized frame. `write_file` shares the same chunking helper.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix `exec()` raising `UnicodeDecodeError` when a command writes non-UTF-8 bytes to stdout/stderr. Invalid bytes are now replaced with the unicode replacement character (matching Inspect's other sandboxes) instead of aborting the call.
 - Propagate the caller's context into the pod-operation worker thread to ensure that Inspect sandbox config overrides are honoured.
 - Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).
 - Fix `exec(input=...)` failing with "Connection reset by peer" for large inputs (e.g. injecting a ~28 MiB binary): stdin is now written to the pod in ≤1 MiB WebSocket frames instead of a single oversized frame. `write_file` shares the same chunking helper.
--- a/src/k8s_sandbox/_pod/buffer.py
+++ b/src/k8s_sandbox/_pod/buffer.py
@@ -4,9 +4,9 @@ class LimitedBuffer:
 
     Once the buffer is full, `truncated` is set and further appends are ignored.
 
-    The buffer can be converted to a string (utf-8). Invalid utf-8 bytes are replaced
-    with the unicode replacement character; an incomplete trailing character left by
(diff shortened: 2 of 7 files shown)


src/k8s_sandbox/_pod/buffer.py
```diff
@@ -22,11 +23,11 @@ def append(self, data: bytes) -> None:
...
     def __str__(self) -> str:
-        # If we're truncated the data, there may be an incomplete character right at the
-        # end of the buffer.
         try:
             return self._buffer.decode("utf-8", errors="strict")
         except UnicodeDecodeError as e:
+            # Drop an incomplete character left at the very end by truncation; replace
+            # any other invalid bytes (e.g. binary subprocess output) rather than raise.
             if self.truncated and e.end == len(self._buffer):
                 return self._buffer[: e.start].decode("utf-8", errors="strict")
```
art-dsit: I think we also need this because the truncation happens at byte level and could chop off the end of UTF-8 character
```suggestion
                return self._buffer[: e.start].decode("utf-8", errors="replace")
```

src/k8s_sandbox/_pod/execute.py
```diff
@@ -193,17 +193,16 @@ def _check_for_runuser_error(self, stderr: str, user: str) -> None:
...
-        # to raise a UnicodeDecodeError if we encounter one, so errors="strict".
-        # Assumption: individual frames are valid utf-8 (i.e. characters are not split
-        # across frames).
-        decoded = frame.decode("utf-8", errors="strict")
-        split_frame = re.split(COMPLETED_SENTINEL_PATTERN, decoded)
-        if len(split_frame) == 1:
+        # The sentinel is ASCII, so decode-with-replace detects it without raising on
+        # binary subprocess output. Only raw bytes pass through, so the replacement
+        # chars can't reach the caller. Assumption: sentinel isn't split across frames.
+        probe = frame.decode("utf-8", errors="replace")
```
art-dsit: I think it would be better to decode as Latin1 (ISO-8859-1) and then re-encode, because of Latin1's ability to roundtrip bytes to characters losslessly

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix `exec()` raising `UnicodeDecodeError` when a command writes non-UTF-8 bytes to stdout/stderr. Invalid bytes are now replaced with the unicode replacement character (matching Inspect's other sandboxes) instead of aborting the call.
```
art-dsit: > Fix `exec()` raising `UnicodeDecodeError` when a command writes non-UTF-8 bytes to stdout/stderr.

Adhering to the spec is not something that you need to fix; we are simply changing the behaviour here

> matching Inspect's other sandboxes

this is too strong an assertion at the moment

src/k8s_sandbox/_pod/buffer.py
```diff
@@ -22,11 +23,11 @@ def append(self, data: bytes) -> None:
         self._buffer.extend(data[:remaining_space])
 
     def __str__(self) -> str:
-        # If we're truncated the data, there may be an incomplete character right at the
-        # end of the buffer.
         try:
             return self._buffer.decode("utf-8", errors="strict")
         except UnicodeDecodeError as e:
+            # Drop an incomplete character left at the very end by truncation; replace
```
dpolatajko-aisi: we could simplify this function to a single line if we dropped this check, and accepted that there might sometimes be a trailing replacement character? but I don't know the nuances of why this check is here, @art-dsit wdyt?

art-dsit: yes we can simplify this; we're accepting replacement characters in all cases now, if I understand correctly

src/k8s_sandbox/_pod/buffer.py
dpolatajko-aisi: these are equivalent right? if the start of the invalid sequence is at e.start, then strict and replace should decode the string up to that point `self._buffer[: e.start]` the same way

CHANGELOG.md
dpolatajko-aisi: i was going to also open a PR against inspect_ai to update the sandbox spec, based on your discussion with JJ. worth doing?

art-dsit: yes I think we should update the spec to say it's implementation-dependent whether it does a unicode replace, or throws (at the moment you CAN throw, but it doesn't say you HAVE to)

dpolatajko-aisi: see here: https://github.com/UKGovernmentBEIS/inspect_ai/pull/4313

test/k8s_sandbox/pod/test_execute.py
```diff
@@ -162,6 +162,44 @@ def test_broken_pipe_without_sentinel_raises(self) -> None:
...
+    def test_binary_byte_with_sentinel_in_same_frame(self) -> None:
+        ws = _make_ws_client(
+            stdout_frames=[b"before \xbb after<completed-sentinel-value-0>"],
+        )
+
+        executor = ExecuteOperation(MagicMock())
+        result = executor._handle_shell_output(ws, user=None, timeout=None)
+
+        assert result.returncode == 0
+        assert result.stdout == "before � after"
```
art-dsit: I wonder if we could hex-encode 'replacement character' rather than embedding it because it looks otherwise like something corrupted our source files

art-dsit:
Do these changes have any effect on read_file or is it strictly only used by exec?

dpolatajko-aisi:
> Do these changes have any effect on read_file or is it strictly only used by exec?

the change to the `__str__` method can impact reads - but afaict the impact is very limited - the only case it makes a difference is when stderr contains non-utf-8 characters during exception handling, in which case the code would no longer mask the root exception with `UnicodeDecodeError` but instead would re-raise the original exception, which seems preferable to me. I could add a unit test which demonstrates that case?

art-dsit (approved):
LGTM, tests are passing

PR #211, opened 2026-06-23
Fix/210 bypass pod deserialization

### LINKED ISSUE #210: scaling bottlenecked by kubernetes client deserialization
I've been trying to scale to ~hundreds of concurrent k8 clusters so I run evals quickly. 

One of the primary bottlenecks seems to be
- `check_for_pod_restart` is called frequently
- this calls `read_namespaced_pod()` which is badly written and locks up with high concurrency (this is a known and long lasting problem, see https://github.com/kubernetes-client/python/issues/2284) 
- this causes the whole eval to grind to a halt with TimeoutErrors

It does seem to be possible to work around by using `read_namespaced_pod(..., _preload_content=False)` and parsing the raw json it returns. That's not a particularly nice solution, but it did cause much better performance when I tested with it. Another possible solution would be calling `check_for_pod_restart` much less frequently (although I do think the issue comes up in some other places).

nix-aisi:
The kubernetes Python client deserializes every API response into its generated model classes via ApiClient.__deserialize. That path is slow and effectively single-threaded, so it becomes a serialization bottleneck under high concurrency: the frequently-called check_for_pod_restart starts timing out and grinds many-cluster evals to a halt
(see https://github.com/kubernetes-client/python/issues/2284).

Add a central k8s_sandbox._pod.snapshot module that reads pods with _preload_content=False and parses only the handful of fields we use out of the raw JSON into an immutable PodSnapshot.

Fixes https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/issues/210

Will be running an eval using this fork tonight, can report back on if seemed to work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Propagate the caller's context into the pod-operation worker thread to ensure that Inspect sandbox config overrides are honoured.
 - Raise typed `PodReplacedError` / `ContainerRestartedError` (instead of `RuntimeError`) when a pod operation detects the pod has been replaced or its container restarted, and refresh the cached pod identity so subsequent operations target the new pod instead of looping against a stale UID. `restarted_container_behavior="warn"` now also refreshes (previously left the cached UID stale).
 - Fix `exec(input=...)` failing with "Connection reset by peer" for large inputs (e.g. injecting a ~28 MiB binary): stdin is now written to the pod in ≤1 MiB WebSocket frames instead of a single oversized frame. `write_file` shares the same chunking helper.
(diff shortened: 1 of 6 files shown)


dpolatajko-aisi (approved):
thanks for the pr, lgtm - did a little benchmarking and agree this is a significant speedup. i'll note for posterity that this does introduce a new coupling to the schema of the pod status object, but that api is stable and entirely justified by the perf increase imo

nix-aisi:
This does seem to be working in my eval run (though lots of moving pieces, I don't have a super clean A/B benchmark result). Pretty sure we want something fixing this to be merged even if not confident on the details.

nix-aisi:
Nice, thanks for the review! I'm happy for this to get merged as is if you are. I'm not a collaborator on the github so you'll need to merge

dpolatajko-aisi:
> I don't have a super clean A/B benchmark result

I benchmarked only the deserialisation step and found it to be ~70x quicker in a single thread - as you say, there's lots of other stuff going on in context but combined with your improved eval performance i'm happy

PR #216, opened 2026-07-04
compose converter: support security_opt (seccomp) and ignore memswap_limit

jjallaire:
## Summary
The compose→Helm converter (`k8s_sandbox/compose/_converter.py`, `_ServiceConverter._convert_service`)
pops each supported service key and then raises `Unsupported key(s) in 'service'` for anything left
over. Two Docker Compose fields have no handler and so fail conversion, even though both have clean
k8s semantics:

- **`security_opt: ["seccomp=<path>"]`** → a `Localhost` `seccompProfile` in the pod's
  `securityContext`. This is the standard k8s mechanism for running a workload under a custom
  seccomp profile.
- **`memswap_limit`** → k8s has no per-pod swap-limit field, and nodes run with swap off by default,
  so the "disable swap" intent of `memswap_limit == mem_limit` is already the ambient behavior on
  k8s. Ignore it with an info log (matching how the converter already treats `init`/`expose`).

Both changes are **purely additive**: they add handling for keys the converter currently rejects, so
no config that converts today changes behavior.

## Motivation
Running an exploitation/security-eval workload faithfully requires **ASLR disabled in the target
container**. The portable, per-container way to do this is `setarch -R <cmd>`
(`personality(ADDR_NO_RANDOMIZE)`), inherited across fork+exec — no host/node sysctl needed. But the
default docker/k8s seccomp profiles **filter `personality(0x40000)`**, silently defeating `setarch -R`.
The fix is a scoped seccomp profile (docker default + `personality` allowed), expressed in compose as
`security_opt: ["seccomp=<path>"]`. On docker this Just Works; on k8s the ComposeConfig currently
can't be converted at all because the converter rejects `security_opt`. This PR closes that gap so the
same ComposeConfig is portable across the docker and k8s providers.

`memswap_limit` shows up alongside it (setting `memswap_limit == mem_limit` to force clean OOM-kills
instead of host swap on docker); it's handled here so the same config converts cleanly rather than
requiring provider-specific config surgery.

## Behavior
- `security_opt: ["seccomp=/path/to/profile.json"]` →
  ```yaml
  securityContext:
    seccompProfile:
      type: Localhost
      localhostProfile: /path/to/profile.json
  ```
  merged with any `securityContext` already produced from `user:`. (Note: `localhostProfile` is
  resolved by the kubelet relative to its seccomp root — the profile must be present on nodes; that's
  a deployment prerequisite, same as any Localhost seccomp profile, and out of scope for the
  converter.)
- Non-`seccomp=` `security_opt` entries (e.g. `apparmor=`, `no-new-privileges`): rejected with an
  explicit `Unsupported 'security_opt' entries: […]` error. We only map `seccomp=`; we don't silently
  swallow other options (which would let a workload believe a security control is applied when it isn't).
- `memswap_limit`: popped + info-logged, not mapped.

## Testing
- Added: a service with `security_opt: ["seccomp=<path>"]` converts and emits the expected
  `securityContext.seccompProfile`; merges with a `user:`-derived `securityContext`; the emitted
  values validate against the chart's `values.schema.json`; a non-`seccomp=` entry raises (both alone
  and alongside a valid `seccomp=` entry); `memswap_limit` is ignored (info-logged, not rejected).
- Existing converter tests unchanged/passing (additive change).

## Notes / alternatives considered
- Could route via the `x-inspect_k8s_sandbox` extension instead of the native `security_opt` field,
  but the extension currently only supports `resources`, and `security_opt` is a first-class compose
  field with an obvious k8s target, so handling it natively is cleaner and more discoverable.
- Scope kept minimal: only `seccomp=` (the portable, widely-useful case). Other `security_opt`
  values can be added later if there's demand.

## Interim usage (downstream)
Until this lands, downstream applies the same mapping via a small, feature-detected monkeypatch of
`_ServiceConverter._convert_service` (pop `security_opt`/`memswap_limit` before the leftover guard,
inject `seccompProfile` after). The patch here supersedes that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- Compose to HELM: Map `security_opt: ["seccomp=<path>"]` to a `Localhost` `seccompProfile` in the pod's `securityContext` (merged with any `user:`-derived context; non-`seccomp=` entries are rejected).
+- Compose to HELM: Ignore `memswap_limit` with an info log, since Kubernetes has no per-pod swap knob and nodes run with swap off by default.
+
 ## 2026-06-25 0.6.1
 
 - no changes - version bump only
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,7 @@
 
 ## Unreleased
 
-- Compose to HELM: Map a `security_opt` seccomp entry (`seccomp=<value>` or `seccomp:<value>`) to a `seccompProfile` in the pod's `securityContext`, merged with any `user:`-derived context. `unconfined`/`builtin` map to the `Unconfined`/`RuntimeDefault` profile types; other values are treated as `Localhost` profile paths (validated as relative, non-empty, no `..`). Non-seccomp entries (e.g. `apparmor=`) are rejected.
-- Compose to HELM: Ignore `memswap_limit` with an info log, since Kubernetes exposes no Compose-equivalent per-container swap limit.
+- Compose to HELM: Support the `security_opt` seccomp option (mapped to a pod `seccompProfile`) and ignore the unsupported `memswap_limit`. See [Compose to Helm](https://k8s-sandbox.aisi.org.uk/helm/compose-to-helm/) for details.
 
 ## 2026-06-25 0.6.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## Unreleased
 
-- Compose to HELM: Map `security_opt: ["seccomp=<path>"]` to a `Localhost` `seccompProfile` in the pod's `securityContext` (merged with any `user:`-derived context; non-`seccomp=` entries are rejected).
-- Compose to HELM: Ignore `memswap_limit` with an info log, since Kubernetes has no per-pod swap knob and nodes run with swap off by default.
+- Compose to HELM: Map a `security_opt` seccomp entry (`seccomp=<value>` or `seccomp:<value>`) to a `seccompProfile` in the pod's `securityContext`, merged with any `user:`-derived context. `unconfined`/`builtin` map to the `Unconfined`/`RuntimeDefault` profile types; other values are treated as `Localhost` profile paths (validated as relative, non-empty, no `..`). Non-seccomp entries (e.g. `apparmor=`) are rejected.
+- Compose to HELM: Ignore `memswap_limit` with an info log, since Kubernetes exposes no Compose-equivalent per-container swap limit.
 
 ## 2026-06-25 0.6.1
 
--- a/docs/docs/helm/compose-to-helm.md
+++ b/docs/docs/helm/compose-to-helm.md
@@ -182,11 +182,19 @@ Notes:
 - Non-seccomp entries (e.g. `apparmor=...`, `no-new-privileges`) have no mapping here and
   are rejected rather than silently dropped, so a workload can't believe a security
   control is applied when it isn't.
+- **Runtime caveat (gVisor).** A seccomp profile only changes which syscalls the kernel
+  *allows*; it can't add syscalls the runtime doesn't implement. The chart defaults to the
+  `gvisor` runtime, whose `personality()` does not support `ADDR_NO_RANDOMIZE`. So a
+  profile that permits `personality` in order to run `setarch -R` (the portable way to
+  disable ASLR, e.g. for exploitation workloads) has no effect under gVisor — `setarch -R`
+  fails with `EINVAL` regardless of the profile. Set `runtime: runc` on the service
+  (mapped to `runtimeClassName: runc`) for those workloads.
 
 ```yaml
 services:
   my-service:
     image: ubuntu
+    runtime: runc  # gVisor (the default) can't disable ASLR; see the runtime caveat above
     security_opt:
       - seccomp=profiles/no-aslr.json
 ```
--- a/docs/docs/helm/compose-to-helm.md
+++ b/docs/docs/helm/compose-to-helm.md
@@ -161,10 +161,10 @@ A `security_opt` **seccomp** entry (`seccomp=<value>` or `seccomp:<value>`) is c
 to a `seccompProfile` in the pod's `securityContext` (merged with any context derived
 from `user`). Docker's special values map to Kubernetes profile types:
 
-| Compose value | Kubernetes `seccompProfile` |
-| --- | --- |
-| `seccomp=unconfined` | `{type: Unconfined}` |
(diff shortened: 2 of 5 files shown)


RoshDSIT:
LGTM - a few non-blocking notes below:
- ignoring `memswap` would be fine for nodes with swap off, but if the cluster has swap enabled and if the docker container wants swap off, this might change the intended behaviour - maybe worth a note in the docs.
- where a `LocalhostProfile` is created, it might be worth a logging line, e.g. "_must be pre-staged on all nodes at /var/lib/kubelet/seccomp/<path>_" -given that this isn't surfaced at conversion time and fails only at pod launch.
- nits: 
  - tests missing for colon-separator for `unconfined` and `builtin`, and for the scalar security opt
  - the error message only says that `"="` is accepted, but doesn't say that `":"` is accepted
- checked `seccomp=builtin` on docker directly - it's accepted, so this looks like a faithful mapping across (though `builtin` isn't in [docker's public docs](https://docs.docker.com/engine/security/seccomp/))

jjallaire:
> LGTM - a few non-blocking notes below:
> 
> * ignoring `memswap` would be fine for nodes with swap off, but if the cluster has swap enabled and if the docker container wants swap off, this might change the intended behaviour - maybe worth a note in the docs.
> * where a `LocalhostProfile` is created, it might be worth a logging line, e.g. "_must be pre-staged on all nodes at /var/lib/kubelet/seccomp/_" -given that this isn't surfaced at conversion time and fails only at pod launch.
> * nits:
>   
>   * tests missing for colon-separator for `unconfined` and `builtin`, and for the scalar security opt
>   * the error message only says that `"="` is accepted, but doesn't say that `":"` is accepted
> * checked `seccomp=builtin` on docker directly - it's accepted, so this looks like a faithful mapping across (though `builtin` isn't in [docker's public docs](https://docs.docker.com/engine/security/seccomp/))

Thx! Feedback addressed here: https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/pull/216/changes/98d92881c81df2a9cdf20a9568331ff88289c22d

docs/docs/helm/compose-to-helm.md
```diff
@@ -144,3 +144,49 @@ This extension exists because Docker Compose cannot express many Kubernetes
 resources. It covers CPU and memory requests/limits (via `cpus`/`mem_limit` and
 `deploy.resources`), but has no concept of a disk (`ephemeral-storage`) request or
 limit, nor of resources such as `hugepages-*`.
+
+### Swap (`memswap_limit`)
+
+`memswap_limit` is ignored (with an info-level log) because Kubernetes has no
+Compose-equivalent per-container swap limit. On typical clusters, swap is disabled on
+nodes, so this is a no-op. However, if your cluster enables swap (via the Kubernetes
+[NodeSwap](https://kubernetes.io/docs/concepts/architecture/nodes/#swap-memory)
```
art-dsit: I think this is https://kubernetes.io/docs/concepts/cluster-administration/swap-memory-management/

src/k8s_sandbox/compose/_converter.py
```diff
@@ -585,6 +619,36 @@ def parse_int(value: str) -> int:
...
+            raise ComposeConverterError(
+                f"Invalid seccomp profile in 'security_opt': '{value}'. Expected "
+                f"'unconfined', 'builtin', or a relative profile path (no absolute "
+                f"paths or '..'). {self.context}"
+            )
+        # Unlike the built-in profile types, a Localhost profile file is not shipped by
+        # the converter: it must already exist on every node. This isn't verified here,
+        # so a missing file surfaces only as a pod launch failure, not a conversion
+        # error.
+        logger.warning(
```
art-dsit: I'm not sure it's helpful to warn every time, even when things are going to work. `info` seems like it would be fine (I realise you put this in as a PR comment response - sorry!)

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+- Compose to HELM: Map a `security_opt` seccomp entry (`seccomp=<value>` or `seccomp:<value>`) to a `seccompProfile` in the pod's `securityContext`, merged with any `user:`-derived context. `unconfined`/`builtin` map to the `Unconfined`/`RuntimeDefault` profile types; other values are treated as `Localhost` profile paths (validated as relative, non-empty, no `..`). Non-seccomp entries (e.g. `apparmor=`) are rejected.
```
art-dsit: I think this can be terser and just signpost the docs

src/k8s_sandbox/compose/_converter.py
```diff
@@ -254,6 +254,40 @@ def _convert_service(self, src: dict[str, Any]) -> dict[str, Any]:
         _transform(
             src, "user", result, "securityContext", self._user_to_security_context
         )
+        # security_opt: map a seccomp entry (`seccomp=<value>` or `seccomp:<value>`) to
+        # a seccompProfile in securityContext (which `user` above may also populate).
+        # Any other security_opt entries (e.g. `apparmor=`, `no-new-privileges`) have no
+        # k8s mapping here and are rejected rather than silently dropped, so a workload
+        # can't believe a security control is applied when it isn't.
```
art-dsit: I think the part of this comment `Any other security_opt entries` onwards can be dropped, it's obvious from the code

docs/docs/helm/compose-to-helm.md
```diff
@@ -144,3 +144,49 @@ This extension exists because Docker Compose cannot express many Kubernetes
...
+
+A `security_opt` **seccomp** entry (`seccomp=<value>` or `seccomp:<value>`) is converted
+to a `seccompProfile` in the pod's `securityContext` (merged with any context derived
+from `user`). Docker's special values map to Kubernetes profile types:
+
+| Compose value | Kubernetes `seccompProfile` |
+| --- | --- |
+| `seccomp=unconfined` | `{type: Unconfined}` |
+| `seccomp=builtin` | `{type: RuntimeDefault}` |
+| `seccomp=<relative/path.json>` | `{type: Localhost, localhostProfile: <relative/path.json>}` |
```
art-dsit, line 168: Is it possible to fix this table so it's nicely aligned?

art-dsit (commented):
A few comments, plus a doc warning + extra test in https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/pull/218 which targets this PR's branch

docs/docs/helm/compose-to-helm.md
jjallaire:
```suggestion
[NodeSwap](https://kubernetes.io/docs/concepts/cluster-administration/swap-memory-management/)
```

art-dsit (approved):

PR #217, opened 2026-07-06
Bump inspect-ai floor to >=0.3.236

art-dsit:
## Summary

Bumps the `inspect-ai` dependency floor from `>=0.3.161` to `>=0.3.236`, and enlarges the `self_check` fixture pods from 128Mi to 256Mi so the suite still passes against current `inspect-ai`.

## Why the floor bump

`0.3.236` is the first `inspect-ai` release whose `ComposeService` model accepts `security_opt` (and `0.3.233` added `memswap_limit`). The k8s sandbox routes `("k8s", "compose.yaml")` through `parse_compose_yaml` → `ComposeConfig` *before* the compose→Helm converter runs, so on older `inspect-ai` those keys are rejected as unknown fields and never reach the converter. This floor is a prerequisite for #216 (compose `security_opt`/`memswap_limit` support) to work through the standard compose entrypoint.

## Why the fixture memory bump

Newer `inspect-ai` `self_check` raised its large-payload stress test from 1 MiB to 50 MiB (exec stdin and read/write), with the comment *"The k8s sandbox blew up at 28 MiB."* Under gVisor, a 50 MiB payload OOM-kills the 128Mi `self_check` fixture pods — `wc` reads 0 bytes, or the exec hits its 30s timeout. This is a fixture-sizing issue, not a transport bug (a 512Mi pod handles 50 MiB fine). Bumping the fixture pods to 256Mi resolves it.

## Testing

- Both `test_self_check_k8s_default_root` and `test_self_check_k8s_non_root` pass on `inspect-ai` 0.3.244 with the 256Mi fixture (verified on minikube).
- Full `req_k8s` suite otherwise green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## Unreleased
 
-- Require `inspect-ai>=0.3.236`.
 - The package and bundled `agent-env` chart versions are now unified, both jumping to
   `0.13.0` (intervening numbers are unused).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Require `inspect-ai>=0.3.236`.
 - The package and bundled `agent-env` chart versions are now unified, both jumping to
   `0.13.0` (intervening numbers are unused).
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Require `inspect-ai>=0.3.236`.
```
art-dsit:
```suggestion
```

no need for this trivia

dpolatajko-aisi (approved):
lgtm

PR #220, opened 2026-07-16
fix: omit successfully completed init containers from pod diagnostics

RoshDSIT:
Follow-up to #202:

On any install failure of a chart that uses init containers, the pod diagnostics added in #202 include a line like `init container 'setup': terminated Completed (exit code 0)`. That is an init container's normal resting state after it has done its job, so the line is noise at the exact moment the output needs to be signal, and for a pod whose containers are otherwise healthy it can be the only line in the summary.

`_describe_container` now skips init containers that terminated with exit code 0. Failed init containers are still surfaced, as are all app container states and Warning events. One deliberate consequence: an init container that crash-looped a few times before eventually succeeding is also omitted, since it is not the cause of the current failure and its churn still shows up in the BackOff events.

Tested with two new unit cases (a completed init container alongside a crashing app container, and the no-op case where the completed init container would have been the only output) and end to end on a minikube cluster, where the same deliberately failing pod produced the noise line before the change and only the real failure afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- On Helm install failure, the raised error now includes pod diagnostics.
 - Compose to HELM: Support the `security_opt` seccomp option (mapped to a pod `seccompProfile`) and ignore the unsupported `memswap_limit`. See [Compose to Helm](https://k8s-sandbox.aisi.org.uk/helm/compose-to-helm/) for details.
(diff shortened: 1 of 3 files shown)


art-dsit:
I'm not sure it's noise; I actually think it's relevant information that a given init container wasn't part of the problem

RoshDSIT:
> I'm not sure it's noise; I actually think it's relevant information that a given init container wasn't part of the problem

That makes sense - I'll close this and make a PR only including the updated changelog

PR #221, opened 2026-07-16
docs: changelog entry for pod diagnostics

RoshDSIT:
Adds the missing Unreleased entry for the pod diagnostics introduced in #202.


art-dsit (approved):
Personally I would not have bothered adding a changelog entry for this because the change was just a quality of life improvement rather than something that might require end-user action, but happy to merge this anyway

PR #226, opened 2026-07-26
fix: verify `write_file` byte count to avoid silent truncation

### LINKED ISSUE #225: `write_file()` can silently write an empty/truncated file with a success return code
Sometimes `write_file()` lands an empty file on the pod (in our case the copied verifier test.sh) while returning success. Downstream this looks baffling: the empty script runs as a no-op, writes no result, and fails much later -- with no error at the write itself. It happens on a subset of tasks and is worse under concurrency                                                                                                                   

### Theoretical cause:                                                                                                                                                                                                          

`write_file()` copies by streaming the bytes over exec stdin into: `/bin/sh -c 'mkdir -p <dir> && head -c N > <dst>'`

Because `head` is the shell's last command, the shell execs into it and applies the `> <dst>` redirect first -- so head's stdout points at the file and nothing holds the exec stdout pipe open anymore. Once the exec stdout stream hits `EOF`, the runtime closes the process's stdin while `head` is still reading. `head -c N` treats the resulting short/empty stdin as normal and exits 0, so the write "succeeds" with a truncated file. Nothing verifies the byte count.                                                                                                                                                               

(Confirmed behavior by the experiments: keeping `head` from being the final command, or holding the exec stdout pipe open, both prevent it. The exact runtime component responsible isn't pinned down here; nodes run containerd 2.1.5)

It only hits a subset of tasks because whether it fires is a race between the streamed bytes arriving and the stdin-close: under load/latency the close tends to win (truncated file); otherwise the bytes arrive first and the write is fine.

### Experiments with monkey-patching:

`head -c N > dst` (pristine) = 15/16 fail

`head -c N > dst && :` (trailing no-op, ~0 delay → head not final) = 0/16 fail

`exec 3>&1; head -c N > dst` (head still final, spare stdout fd kept open) = 0/16 fail

  `&& :` fixing it with no added delay rules out timing/delivery as the cure; `exec 3>&1` fixing it while head stays final pins the cause to the stdout-EOF → stdin-close coupling.                                  


### Proposed fix:

Verify the landed byte count so a short write fails loudly and the existing `write_file` retry re-runs it

samorojy:
Fixes #225  

`write_file()` could silently write a truncated/0-byte file while returning success (`head -c N` exits 0 on a short read, and nothing verified the result.)

Solution:
Appends a byte-count check to the write command:
`mkdir -p <dir> && head -c N > <dst> && [ "$(wc -c < <dst>)" -eq N ]`


samorojy:
Following the discussion in the issue, I switched to the `exec 3>&1;`, it fixes the cause rather than the symptom.

It duplicates the exec's stdout onto fd 3, which survives the shell's exec into `head` and so keeps a writer on the pipe for its lifetime. I also added a `req_k8s` test asserting the written byte count, plus a negative control that strips the fix back out and checks the truncation returns, both verified failing without the fix and passing with it.

dpolatajko-aisi:
> Following the discussion in the issue, I switched to the `exec 3>&1;`, it fixes the cause rather than the symptom.
> 
> It duplicates the exec's stdout onto fd 3, which survives the shell's exec into `head` and so keeps a writer on the pipe for its lifetime. I also added a `req_k8s` test asserting the written byte count, plus a negative control that strips the fix back out and checks the truncation returns, both verified failing without the fix and passing with it.

thanks ! i pushed a test fix, since the issue doesn't replicate on gvisor which the tests run against, now looking good to me

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix `write_file()` silently writing a truncated or empty file while reporting success
+  ([#225](https://github.com/UKGovernmentBEIS/inspect_k8s_sandbox/issues/225)).
```
dpolatajko-aisi:
```suggestion
```

dpolatajko-aisi (approved):

samorojy:
🔥, thanks for the quick help :)

Do you have an estimate for when the fix will be released?

PR #227, opened 2026-07-26
fix: report Helm releases that fail to uninstall

sjawhar:
## Problem

`HelmReleaseManager.uninstall_all` gathers the uninstalls with `return_exceptions=True` and never inspects the results, having already cleared `_installed_releases` before awaiting:

```python
tasks = [release.uninstall(quiet=False) for release in self._installed_releases]
# Clear the list before awaiting the tasks to prevent other calls to this method
# from interfering.
self._installed_releases.clear()
await asyncio.gather(*tasks, return_exceptions=True)
```

`helm uninstall` failures do raise (`_helm.uninstall` calls `_raise_runtime_error` when the subprocess fails), so a failed uninstall is swallowed *and* the release has already been dropped from tracking. Nothing retries it, nothing reports it, and `_print_cleanup_instructions` can no longer list it. The release stays installed in the cluster with no record anywhere that it existed.

This is most likely to bite at task cleanup, because `sample_cleanup` returns early for interrupted samples and defers those releases to `uninstall_all`:

```python
# If we were interrupted, wait until the end of the task to cleanup (this
# enables us to show output for the cleanup operation).
if interrupted:
    return
```

So the releases most likely to reach `uninstall_all` are exactly the ones with no other cleanup path.

We hit the consequences of a leaked sandbox on a shared cluster. The pod stays alive, and because sandbox pods can carry `karpenter.sh/do-not-disrupt`, the node cannot be reclaimed. In our case that blocked an unrelated infrastructure change for hours. Our leak had a separate root cause, so this PR is not a fix for that incident; it is the reason we went looking at this function, and the silent-failure path is a real defect on its own.

## Change

Log each failed uninstall with the release name and the `inspect sandbox cleanup k8s <release>` command that removes it.

Behaviour is otherwise unchanged. One failure still does not prevent the other releases being uninstalled, and the list is still cleared before awaiting so concurrent calls cannot interfere. I deliberately did not re-add failed releases to `_installed_releases`, since that would race with the concurrent-call case the existing comment guards against.

## Tests

`test/k8s_sandbox/test_manager.py`, no cluster required.

Verified non-circular: with the source change reverted and the test kept, `test_uninstall_all_reports_a_failed_uninstall` fails on `assert 'bbbbbbbb' in ''`. The empty log is the silent swallow itself.

```
uv run pytest -q -m "not req_k8s"     388 passed, 246 deselected
uv run ruff check . / format --check  All checks passed
uv run mypy .                         Success: no issues found in 84 source files
```

Not run: the `req_k8s` suite, as I do not have a cluster available for this branch.

Written by Claude, acting on behalf of Sami Jawhar.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- A Helm release which fails to uninstall during cleanup is now reported by name, with the `inspect sandbox cleanup k8s <release>` command to remove it. Previously the failure was silent and the release was left installed in the cluster.
 - On Helm install failure, the raised error now includes pod diagnostics.
 - Compose to HELM: Support the `security_opt` seccomp option (mapped to a pod `seccompProfile`) and ignore the unsupported `memswap_limit`. See [Compose to Helm](https://k8s-sandbox.aisi.org.uk/helm/compose-to-helm/) for details.
 - The package and bundled `agent-env` chart versions are now unified, both jumping to
--- a/src/k8s_sandbox/_manager.py
+++ b/src/k8s_sandbox/_manager.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 from contextvars import ContextVar
 
 from rich import box, print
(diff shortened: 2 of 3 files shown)


dpolatajko-aisi:
thanks for the PR - accepting with some additions I made in #229, so I'll close this

PR #228, opened 2026-07-27
fix(network-policy): make sandbox default-deny-ingress actually deny

sjawhar:
The `-sandbox-default-deny-ingress` CiliumNetworkPolicy renders `ingress:` followed by `- {}`. An IngressRule with no selectors matches every source, so that empty rule is allow-all: the policy named for denying ingress permits it instead. Any chart consumer relying on it for pod isolation does not have it.

## The fix

```yaml
  enableDefaultDeny:
    ingress: true
    egress: false
  ingress:
    - fromEntities: []
```

Both parts are load-bearing, and neither is obvious from the field names.

`ingress: []` on its own does not work. Cilium's `Rule` docs state that an omitted or empty top-level `ingress` means the rule does not apply at ingress, and from 1.17 a rule whose `ingress`, `ingressDeny`, `egress` and `egressDeny` are all empty is rejected by the agent. An empty list makes the policy inert rather than strict.

`fromEntities: []` is what selects nothing. Cilium special-cases empty non-nil slices in `SetAggregatedSelectors` and `GetSourceEndpointSelectorsWithRequirements` (`pkg/policy/api/ingress.go`), returning nil selectors, with the comment "explicitly check for empty non-nil slices, it should not result in any identity being selected". That is distinct from an absent field, which yields no L3 constraint, and is why a `toPorts`-only rule allows every source. So the empty list contributes no allow while keeping the stanza non-empty for the validator, and `enableDefaultDeny.ingress` states the intended posture explicitly rather than relying on it being inferred.

## Behaviour preserved

The per-service `-svc-<name>-ingress` policies still render and still allow same-sandbox traffic. Egress policies (`allowDomains`, `allowCIDR`, `allowEntities`) are untouched. Consumers that inject their own ingress allows through `additionalResources` continue to work, since those compose as a union with this policy's deny posture.

## Tests

Render tests added to `test/k8s_sandbox/helm_chart/test_helm_chart.py`, which shells out to `helm template`. They cover the default chart and a `networks`-enabled chart where a service has no `networks` of its own and therefore gets no per-service ingress policy: that second case is where nothing else would establish the deny posture.

`test/k8s_sandbox/test_network_policy.py` was not used, as it is marked `req_k8s` at module level and a render assertion there would be skipped in non-cluster runs.

`uv run pytest test/k8s_sandbox/helm_chart/` passes 34; `uv run mypy .` is clean across 83 files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix the built-in chart's `-sandbox-default-deny-ingress` policy, which allowed all ingress instead of denying it. Ingress to sandbox pods not allowed by another policy is now denied.
 - `inspect sandbox cleanup k8s` (with no release name) now **exits non-zero** if any release fails to uninstall, rather than reporting `Complete.` and exiting 0. Releases which fail to uninstall are named, at end-of-task cleanup too, along with their namespace and the `inspect sandbox cleanup k8s <release>` command to retry them.
 - **BREAKING CHANGE**: Sandbox pods created by the built-in Helm chart no longer mount
   Kubernetes service-account API tokens by default. Set
--- a/src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
+++ b/src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
@@ -129,13 +129,19 @@ metadata:
   labels:
     {{- toYaml $.Values.labels | nindent 4 }}
 spec:
-  description: Default deny ingress. Allow other policies to be more specific.
+  description: |
+    Default deny ingress. The empty fromEntities selector matches no identities;
+    the non-empty ingress rule makes Cilium apply this policy's explicit
+    default-deny posture. Other policies provide permitted ingress.
   endpointSelector:
     matchLabels:
       io.kubernetes.pod.namespace: {{ .Release.Namespace }}
       {{- include "agentEnv.selectorLabels" $ | nindent 6 }}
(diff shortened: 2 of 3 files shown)


art-dsit:
I can't reproduce this. I added a commit 101aa75 with a test that I argue would fail if the PR description was correct, but it passes on main

sjawhar:
Claude here, on Sami Jawhar's behalf. You were right, and I apologise for the noise.

I re-ran your test shape on our staging EKS cluster (Cilium 1.18.4) as a causal A/B/A experiment in a disposable namespace, five attempts per phase:

| Phase | Policy | Result |
|---|---|---|
| A | none | HTTP 200, 5/5 |
| B | `ingress: [{}]` | curl exit 28 (timeout), 5/5 |
| A' | policy removed | HTTP 200, 5/5 |
| control | same target, explicit `fromEndpoints` allow for the attacker's labels | HTTP 200, 5/5 |

Cilium's endpoint state matched: no policy `policy_enabled: none`; with `ingress: [{}]` `policy_enabled: ingress` and the attacker identity absent from the allowed set; with the explicit allow, present.

On the representation question: `ingress: []` is rejected by Cilium's validator ("rule must have at least one of Ingress, IngressDeny, Egress, EgressDeny") and an omitted `ingress` is rejected at Kubernetes admission, so the one-empty-rule form is the only valid shape and it enforces default deny.

Our premise was wrong. Closing this. Sorry for the time it cost you, and thanks for writing the e2e test rather than just taking our word for it - it would be worth landing on its own as a regression guard.

PR #229, opened 2026-07-28
fix: report all Helm cleanup failures

dpolatajko-aisi:
See #227 for context - @sjawhar identified and fixed a bug related to Helm release cleanup, but the same bug existed in multiple places, so I applied the fix more broadly here. Supersedes #227


RoshDSIT:
Nice - this looks good. The changelog entry could be slightly louder about the new exit code (nit).

nit: Claude suggesting that the logging msg should include the `Release` `_namespace`/`_context_name`, pointing to a potential failure mode which is already documented in `docs/docs/tips/configuration.md:82)`: 
- user runs an eval with e.g. `K8sSandboxEnvironmentConfig(context="minikube")`; 
- uninstall fails;
- `uninstall_unmanaged_release` hardcodes `context_name=None`
- `helm uninstall` with `--ignore-not-found` would attempt cleanup against the wrong cluster
It feels like a minimal addition in logging msg.

note: Claude also noting that Ctrl+C during cleanup will leave the orphaned releases, which feels like it's to be expected, but logging could be improved in a follow up.

Did the `req_k8s` tests run?

RoshDSIT (approved):

PR #231, opened 2026-08-06
fix: reject conflicting pod operation limits

sjawhar:
## Summary

`PodOpExecutor.get_instance(max_pod_ops=...)` previously ignored a supplied limit after a parameterless call had created the singleton. The executor now raises `ValueError` when a later supplied limit conflicts with the singleton capacity.

This was exposed on 2026-08-06 in a 400-sandbox Kubernetes eval set: `pod-op` was capped at 128 while 354 sandboxes contended, and 414 of 621 resolved samples errored with `MCPEndpointsUnreachableError`. Raising the live limit to 384 reduced the failure rate from 70% to 11%.

## Test

Adds a regression test for a parameterless initialization followed by a conflicting `max_pod_ops` value. It fails against the previous silent fallback.

## Agent review

- `uv run --with ruff==0.9.6 ruff check .`
- `uv run --with ruff==0.9.6 ruff format --check .`
- `uv run --extra dev mypy .`
- `uv run --extra dev pytest -m "not req_k8s" -q` (405 passed, 247 deselected)
- `uv run --extra dev pre-commit run --all-files`
- `make check` and `make test` are not available: this repository has no Makefile.

Authored with AI assistance by Claude.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Raise an error when a conflicting `max_pod_ops` setting would otherwise be ignored.
+
 ## 2026-08-12 0.13.0
 
 - Fix `write_file()` silently writing a truncated or empty file while reporting success
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Raise an error when a conflicting `max_pod_ops` setting would otherwise be ignored.
 - `inspect sandbox cleanup k8s` (with no release name) now **exits non-zero** if any release fails to uninstall, rather than reporting `Complete.` and exiting 0. Releases which fail to uninstall are named, at end-of-task cleanup too, along with their namespace and the `inspect sandbox cleanup k8s <release>` command to retry them.
 - **BREAKING CHANGE**: Sandbox pods created by the built-in Helm chart no longer mount
   Kubernetes service-account API tokens by default. Set
--- a/src/k8s_sandbox/_pod/executor.py
+++ b/src/k8s_sandbox/_pod/executor.py
@@ -56,14 +56,20 @@ class PodOpExecutor:
         Args:
             max_pod_ops: Maximum number of concurrent pod operations. If provided
                 on the first call, overrides the INSPECT_MAX_POD_OPS env var and
-                the default (cpu_count * 4). Ignored on subsequent calls since the
-                singleton is already created.
+                the default (cpu_count * 4). A later call with a different value
+                raises ValueError rather than silently ignoring the configuration.
 
         This method is async-safe (because it doesn't await anything) but not
         thread-safe.
         """
         if cls._instance is None:
             cls._instance = cls(max_pod_ops=max_pod_ops)
+        elif max_pod_ops is not None and cls._instance._max_workers != max_pod_ops:
+            raise ValueError(
+                "PodOpExecutor is already initialized with "
+                f"max_pod_ops={cls._instance._max_workers}; cannot use "
+                f"max_pod_ops={max_pod_ops}."
+            )
         return cls._instance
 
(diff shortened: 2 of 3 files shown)


sjawhar:
CI on this PR is red from **GitHub Actions infrastructure failures, not test failures**. All seven red checks on run 31117566275 (2026-08-06) failed identically at the action-resolution step, before any code ran:

```
Getting action download info
Failed to resolve action download info. Error: Service Unavailable
Retrying in 12.914 seconds
Failed to resolve action download info. Error: Service Unavailable
##[error]Service Unavailable
```

`test / pytest (helm 3.20.0)` is the one job that got past it and it **passed**. `gh run rerun --failed` needs admin rights on this repo, so this needs a maintainer to re-run before the red is meaningful.

---

Separately, an independent production confirmation of the failure mode this PR guards, from a different production eval set (2026-08-07).

That run set `INSPECT_MAX_POD_OPS=256` and got a resolved `max_sandboxes=400` — **1.56x oversubscribed**. Exec latency steps at exactly the pool boundary. Full census of all 762 `.eval` archives (7,614 samples, 170,165 sandbox events, 0 parse failures), binned by mean in-flight sandbox events per 30 s window:

| mean in-flight | vs 256 pool | windows | exec N | exec p50 |
|---|---|---:|---:|---:|
| 0–64 | 0.12x | 70 | 15,597 | **1.86 s** |
| 64–128 | 0.38x | 57 | 17,804 | 2.40 s |
| 128–192 | 0.62x | 12 | 9,467 | 3.86 s |
| 192–256 | 0.88x | 11 | 14,994 | 4.58 s |
| 256–320 | **1.12x** | 35 | 34,055 | **10.43 s** |
| 320–384 | 1.38x | 42 | 43,769 | 9.53 s |
| 384–512 | 1.75x | 9 | 10,714 | 9.34 s |

Split at the pool size: 148 windows below → median exec p50 **2.39 s**; 86 windows at or above → **9.69 s**. A **4.1x** step in the *median*, for commands that are mostly sub-second `curl`/`grep`.

Reproduced the mechanism directly against live sandbox pods, driving the real `ExecuteOperation` through a private `ThreadPoolExecutor` (shared singleton untouched):

```
pool=2, demand=4, sleep=5s  -> wall 10.11s (2 waves), distinct threads used = 2
pool=4, demand=4,  ratio 1.0x -> wall 3.06s, queue_p50 0.00s
pool=4, demand=8,  ratio 2.0x -> wall 6.11s, queue_p50 1.52s
pool=4, demand=12, ratio 3.0x -> wall 9.17s, queue_p50 3.05s
```

Wall time tracks `ceil(demand/pool) x service_time` exactly — each exec pins a worker for the **whole command duration** (`_pod/execute.py:_handle_shell_output` blocks in `ws_client.update(timeout=None)`), so oversubscription converts directly into queue latency.

That matches the 400-sandbox/128-pool incident in the PR description, and the corresponding MCP failures: on this run every MCP RPC is itself a sandbox exec (`inspect_ai/tool/_mcp/_sandbox.py` -> `SandboxJSONRPCTransport` -> `pod.exec` -> `PodOpExecutor`), so MCP calls queue behind agent bash calls in the same pool. First `timed out awaiting tools/call` appeared at 18:27:58, after the run crossed the 400-in-flight band at 18:18:30.

One suggestion for scope, offered as a question rather than a request: this PR rejects a *conflicting explicit argument*, which catches the "silently ignored parameter" case. The case that bit this run was different — a pool that is internally consistent but simply smaller than the sandbox ceiling it has to serve. Nothing compares `INSPECT_MAX_POD_OPS` against `max_sandboxes`. Would a warning when the pool is smaller than the caller's sandbox concurrency be in scope here, or is that better placed in the harness that knows the sandbox ceiling (tracked our side as METR/hawk#1235)?

art-dsit (approved):
LGTM, I tested it and it works as expected. Thanks!

PR #232, opened 2026-08-07
perf: skip log formatting when the destination log level is disabled

sjawhar:
## Problem

Every sandbox operation goes through `_log_op` → `inspect_trace_action`, which formats its kwargs into JSON before handing them to Inspect's `trace_action`. `trace_action` writes at `TRACE`, so on a runner logging at `WARNING` every one of those strings is built and immediately discarded.

Found while profiling an eval-set runner that had stopped answering its Inspect control channel. `py-spy` sampling:

```
369  _format_kwargs_as_json  (k8s_sandbox/_logger.py:117)
331  __instancecheck__       (<frozen abc>:117)
167  __init__                (inspect.py:2768)
151  read                    (ssl.py:1140)
 50  update                  (kubernetes/stream/ws_client.py:190)
```

~26% of samples in `_format_kwargs_as_json`, with the abc/inspect machinery it drives accounting for most of the remainder — against ~18% for the actual TLS and WebSocket work.

The process was burning ~1 core of GIL with **31 cores idle** and **325 of its 326 threads asleep**. One thread executes Python at a time, so the asyncio event loop couldn't get enough GIL time to service its own socket. Nothing was wrong with the pod — the loop was simply never free. Adding CPU doesn't help; the GIL is the constraint.

Confirmed on that runner that the work was entirely wasted:

```
logger.isEnabledFor(TRACE) = False
logger.isEnabledFor(DEBUG) = False
logger.isEnabledFor(INFO)  = False
effective level: 30 (WARNING),  root handlers: 0
```

## Change

Skip the whole context manager when `TRACE` is disabled, not just the formatting.

`trace_action` is purely observational — every branch does nothing but `logger.log(TRACE, ...)` and it ends in a bare `raise` — so control flow is identical either way, and skipping it also avoids the uuid, the monotonic timers, and `traceback.format_exc()` on the error path.

| variant | per-op |
|---|---|
| today (format always) | 16.2 µs |
| gate only the formatting | 8.3 µs (−48%) |
| **skip the context manager too** | **1.2 µs (−92%)** |

`log_debug` and `log_trace` get the same treatment. `log_error` and `log_warn` are deliberately left alone — they always emit.

The checks are per-call, so raising the log level at runtime still produces full detail.

## Tests

Four new tests in `test/k8s_sandbox/test_logger.py`. The "was it formatted?" assertion uses an argument whose `__str__` raises, so a regression is a hard failure rather than something a mock might miss:

```python
class _ExplodingArg:
    def __str__(self) -> str:
        raise AssertionError("argument was formatted despite the log level being off")
```

Three assert nothing is formatted when the level is off; the fourth asserts the enabled path **still** formats, so the gating can't silently drop detail when tracing is genuinely on.

Red/green verified: removing the gating fails exactly the three new tests (`3 failed, 10 passed`); restored → `13 passed`. Full suite `408 passed`, ruff and mypy clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Stop formatting log arguments for every pod operation when the destination log level is disabled. Sandbox operations no longer spend ~16us each building trace messages that are then discarded, which on a busy eval-set runner was consuming a full CPU core of Python execution and could leave the run unable to answer `inspect ctl`.
+
 - `inspect sandbox cleanup k8s` (with no release name) now **exits non-zero** if any release fails to uninstall, rather than reporting `Complete.` and exiting 0. Releases which fail to uninstall are named, at end-of-task cleanup too, along with their namespace and the `inspect sandbox cleanup k8s <release>` command to retry them.
 - **BREAKING CHANGE**: Sandbox pods created by the built-in Helm chart no longer mount
   Kubernetes service-account API tokens by default. Set
--- a/src/k8s_sandbox/_logger.py
+++ b/src/k8s_sandbox/_logger.py
@@ -4,6 +4,7 @@ import os
 from contextlib import contextmanager
 from typing import Any, Generator
 
+from inspect_ai._util.logger import TRACE
 from inspect_ai.util import trace_action, trace_message
 
 logger = logging.getLogger(__name__)
@@ -23,6 +24,8 @@ def log_trace(message: str, **kwargs: Any) -> None:
           they exceed DEFAULT_ARG_TRUNCATION_THRESHOLD (which can be overridden with env
           var INSPECT_K8S_LOG_TRUNCATION_THRESHOLD).
     """
+    if not logger.isEnabledFor(TRACE):
+        return
     formatted = format_log_message(message, **kwargs)
     trace_message(logger, category="K8s", message=formatted)
 
@@ -36,6 +39,8 @@ def log_debug(message: str, **kwargs: Any) -> None:
           they exceed DEFAULT_ARG_TRUNCATION_THRESHOLD (which can be overridden with env
           var INSPECT_K8S_LOG_TRUNCATION_THRESHOLD).
     """
+    if not logger.isEnabledFor(logging.DEBUG):
+        return
     formatted = format_log_message(message, **kwargs)
     logger.debug(f"K8s: {formatted}")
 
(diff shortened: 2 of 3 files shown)


sjawhar:
Independent confirmation of the perf claim, measured in a different production runner (2026-08-07, `inspect_k8s_sandbox 0.13.0` without this patch).

Benchmarked in-process in the live runner pod at its real log level, with a realistic `_log_op` kwargs payload (100 KB `ExecResult`, `cmd`, `pod`, `task_name`):

```
deployed log level = WARNING, isEnabledFor(DEBUG) = False

_format_kwargs_as_json (200B result)                 11.7 us/op
_format_kwargs_as_json (10KB result)                 27.8 us/op
_format_kwargs_as_json (100KB result)               146.3 us/op
_format_kwargs_as_json (1MB result)                1365.5 us/op

inspect_trace_action UNGUARDED (100KB)              155.2 us/op
inspect_trace_action guarded-equivalent               8.0 us/op
=> 19.4x cheaper
```

Same direction and rough magnitude as the 16.2 -> 1.2 us in the PR description; the absolute numbers are higher here because the payload is bigger, which is the point — the cost scales with the discarded string.

One extra data point for the "cost scales with payload" argument: `_truncate_arg` does `str(arg)` on the **full** object before truncating to 1000 chars, so the whole `ExecResult` is stringified regardless of the truncation threshold.

```
_truncate_arg(5MB str)                    0.02 ms   (str is str, no copy)
_truncate_arg(ExecResult with 5MB stdout) 11.45 ms  (__repr__ builds the whole thing)
```

That run had `INSPECT_SANDBOX_MAX_EXEC_OUTPUT_SIZE=104857600` (100 MB), so the worst case is bounded only by that limit. Skipping the context manager entirely, as this PR does, avoids it rather than just capping it.

Context on why this mattered operationally there: that runner had `INSPECT_MAX_POD_OPS=256` against a resolved `max_sandboxes=400`, and exec p50 stepped from ~1.85 s to ~9.7 s once mean in-flight sandbox events crossed the pool size (full-census measurement over all 762 `.eval` archives: METR/hawk#930). So the wasted formatting was competing for the GIL with a pod-op pool that was already oversubscribed — consistent with the py-spy picture in the PR description.

Checks are green on this PR; the two neighbouring pod-op PRs (#231, and the fork-only slow-op instrumentation) were both absent from the release that run pinned, which is why none of the intended diagnostics fired: METR/hawk#1235.

art-dsit:
This looks like a fine improvement. I don't think any of the code comments are necessary - the PR description can serve as documentation (if anyone reading in future cares, which seems unlikely to me). Ditto the new test, but happy to keep that if you insist.

I also wouldn't bother putting this in the changelog - we don't want to pollute it with every single change; only changes that a user might need to know about

sjawhar:
Claude, acting on Sami’s behalf: removed the code comments and changelog entry per the review. We kept the test and are happy to remove it too if preferred.

art-dsit (approved):

PR #233, opened 2026-08-10
Harden the CoreDNS sandbox sidecar

oliver-aisi:
Tighten the security of our setup by using a hardened coreDNS sidecar with more restricted permissions. Codex flagged this as a weakness when I was working on unrelated code. However, as noted, this is a breaking change and could cause some annoyance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@
 ## Unreleased
 
 - Fix `write_file()` silently writing a truncated or empty file while reporting success
+- **BREAKING CHANGE**: The CoreDNS sidecar now runs as UID/GID 65532 on a read-only root
+  filesystem with only `NET_BIND_SERVICE`. A custom `corednsImage` must run under that
+  context; set the new `corednsSecurityContext` if it cannot. The default image moves
+  from CoreDNS 1.8.3 to a digest-pinned CoreDNS 1.14.6.
+- **BREAKING CHANGE**: Service names, network names and `additionalDnsRecords` are now
+  validated at install time and rejected with a schema error, rather than producing a
+  release that fails later or resolves unexpectedly.
+- The CoreDNS sidecar no longer serves its `ready` endpoint on port 8181, and refuses
+  queries beyond 1000 concurrent.
 - `inspect sandbox cleanup k8s` (with no release name) now **exits non-zero** if any release fails to uninstall, rather than reporting `Complete.` and exiting 0. Releases which fail to uninstall are named, at end-of-task cleanup too, along with their namespace and the `inspect sandbox cleanup k8s <release>` command to retry them.
 - **BREAKING CHANGE**: Sandbox pods created by the built-in Helm chart no longer mount
   Kubernetes service-account API tokens by default. Set
@@ -17,7 +26,7 @@
 - On Helm install failure, the raised error now includes pod diagnostics.
 - Compose to HELM: Support the `security_opt` seccomp option (mapped to a pod `seccompProfile`) and ignore the unsupported `memswap_limit`. See [Compose to Helm](https://k8s-sandbox.aisi.org.uk/helm/compose-to-helm/) for details.
 - The package and bundled `agent-env` chart versions are now unified, both jumping to
-  `0.13.0` (intervening numbers are unused).
+  `0.14.0` (intervening numbers are unused).
 
 ## 2026-06-25 0.6.1
 
--- a/src/k8s_sandbox/resources/helm/agent-env/README.md
+++ b/src/k8s_sandbox/resources/helm/agent-env/README.md
@@ -1,6 +1,6 @@
 # agent-env
 
-![Version: 0.13.0](https://img.shields.io/badge/Version-0.13.0-informational?style=flat-square)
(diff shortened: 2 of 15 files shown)


src/k8s_sandbox/resources/helm/agent-env/templates/services.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   name: {{ template "agentEnv.fullname" $ -}}-{{ $name }}
   labels:
     {{- include "agentEnv.labels" $ | nindent 4 }}
+    inspect.aisi.org/k8s-sandbox-version: {{ $.Chart.Version | quote }}
```
jon-aisi: nit: the inspect domain is `inspect.aisi.org.uk`, but we've often used `aisi.gov.uk` in k8s labels

jon-aisi: oh, I'm unclear if this typo is in inspect itself already and this just propagates it

test/k8s_sandbox/helm_chart/test_helm_chart.py
```diff
@@ -28,6 +28,12 @@ def test_default_chart(chart_dir: Path) -> None:
         services[0]["spec"]["template"]["spec"]["containers"][0]["image"]
         == "python:3.12-bookworm"
     )
+    assert (
+        services[0]["spec"]["template"]["metadata"]["labels"][
+            "inspect.aisi.org/k8s-sandbox-version"
+        ]
+        == "0.14.0"
```
jon-aisi: this is gonna need updating each release, is it useful enough to be worth it?

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
```
jon-aisi: update to 0.14.0 if we're bumping the version?

src/k8s_sandbox/resources/helm/agent-env/README.md
```diff
@@ -14,7 +14,7 @@
 | annotations | object | `{}` | A dict of annotations to apply to resources within the agent environment. |
 | automountServiceAccountToken | bool | `false` | Whether to mount the selected ServiceAccount's Kubernetes API token in sandbox pods. Keep disabled for cloud workload identity such as IRSA, which injects a provider-specific token separately. |
 | corednsCommand | list | `["/coredns","-conf","/etc/coredns/Corefile"]` | The command to use for the coredns container. |
-| corednsImage | string | `"coredns/coredns:1.8.3"` | The image to use for the coredns container. |
+| corednsImage | string | `"coredns/coredns:1.14.6@sha256:900f9c109f7a33545d3c811516e8376df9019147b750f5ce3e254468769176ea"` | The image to use for the coredns container. Images must be executable as UID/GID 65532 with a read-only root filesystem. |
```
jon-aisi: I believe this is a generated file from `src/k8s_sandbox/resources/helm/agent-env/values.yaml` comments, and this bit
```
Images must be executable as UID/GID 65532 with a read-only root filesystem.
```
isn't in the source file

src/k8s_sandbox/resources/helm/agent-env/values.schema.json
```diff
@@ -96,7 +100,9 @@
             "additionalDnsRecords": {
               "type": "array",
               "items": {
-                "type": "string"
+                "type": "string",
+                "maxLength": 253,
+                "pattern": "^(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\\.)*[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\\.?$"
```
jon-aisi: this doesn't allow underscores, which seems fine & often used by things like ` _dmarc.example.com`?

jon-aisi (commented):
seems like a good change! 

so we're on the same page, my understanding is coredns being messed with can only affect eval integrity, i.e. agents redirecting other pods in that run to unexpected locations, but does not allow breaking out of the sandboxed network restrictions to e.g. gain internet access when it shouldn't, which are enforced by Cilium

I think the only blocking thing is the overly restrictive `additionalDnsRecords` regex

test/k8s_sandbox/helm_chart/test_helm_chart.py
oliver-aisi: I think that we can probably do better with just asserting a minimum version?

oliver-aisi:
> so we're on the same page, my understanding is coredns being messed with can only affect eval integrity, i.e. agents redirecting other pods in that run to unexpected locations, but does not allow breaking out of the sandboxed network restrictions to e.g. gain internet access when it shouldn't, which are enforced by Cilium

I think that this is right, but it generally seems wise to limit the attack surface here lest I am wrong about this.

src/k8s_sandbox/resources/helm/agent-env/values.schema.json
oliver-aisi: have updated to allow underscores!

jon-aisi (approved):

PR #236, opened 2026-08-12
fix: render service `args` with toYaml so compose `command:` survives

### LINKED ISSUE #219: compose `command:` doesn't round-trip to helm args correctly
## Summary

`src/k8s_sandbox/resources/helm/agent-env/templates/services.yaml` renders a service's `args` field without `toYaml`, unlike `command` a few lines above:

```yaml
        {{- if $service.command }}
        command:
          {{- toYaml $service.command | nindent 10 }}
        {{- end }}
        {{- if $service.args }}
        args: {{ $service.args }}
        {{- end }}
```

The compose converter maps compose's `command:` to helm's `args` (and compose's `entrypoint:` to helm's `command`, see `_converter.py`). Since `args` is a list, Go's default stringification renders `args: {{ $service.args }}` as `args: [setarch -R /bin/echo hello]`, which doesn't parse into the intended list of arguments — the container ends up executing something unrelated to the compose `command:` value instead of the process the user asked for.

## Repro

```yaml
# compose.yaml
services:
  default:
    image: ubuntu
    command: ["setarch", "-R", "/bin/echo", "hello"]
```

Convert and render:

```bash
uv run python -c "
from pathlib import Path
import json
from k8s_sandbox.compose._converter import convert_compose_to_helm_values
print(json.dumps(convert_compose_to_helm_values(Path('compose.yaml')), default=str))
" > values.json

helm template test src/k8s_sandbox/resources/helm/agent-env -f values.json
```

The rendered StatefulSet's container has a malformed/incorrect `args` entry rather than `args: ["setarch", "-R", "/bin/echo", "hello"]`.

## Workaround

Using compose's `entrypoint:` instead of `command:` avoids the bug, since `entrypoint` maps to helm's `command` field, which does use `toYaml`.

## Suggested fix

Change:

```yaml
        {{- if $service.args }}
        args: {{ $service.args }}
        {{- end }}
```

to:

```yaml
        {{- if $service.args }}
        args:
          {{- toYaml $service.args | nindent 10 }}
        {{- end }}
```

Found while fact-checking the gVisor/seccomp caveat in `docs/docs/helm/compose-to-helm.md`.

ppcvote:
Closes #219. Thanks @art-dsit for the diagnosis and the repro in the issue; this implements the suggested fix, plus a regression test and a sweep for sibling sites.

## The defect

`templates/services.yaml` rendered `args: {{ $service.args }}`, so Go's default stringification wrote the list as:

```yaml
args: [setarch -R /bin/echo hello]
```

YAML flow sequences are comma-delimited and that output has no commas, so it parses back as a **one-element** sequence holding a single space-joined string. The container execs one argv token instead of the four the caller wrote.

It reaches users through `compose/_converter.py`, which maps compose's `command:` onto Helm's `args`, so any compose file using list-form `command:` was affected. `entrypoint:` was not, because it maps onto Helm's `command` three lines above, which already used `toYaml`.

## The fix

```diff
         {{- if $service.args }}
-        args: {{ $service.args }}
+        args:
+          {{- toYaml $service.args | nindent 10 }}
         {{- end }}
```

This matches the `command:` block immediately above and the initContainer `args` at lines 59-61, which were already correct.

**Swept the rest of the chart for the same shape.** Every other bare interpolation in `templates/` is a scalar (`image`, `imagePullPolicy`, `workingDir`, `port`, `serviceName`, `automountServiceAccountToken`), so this was the only list-valued field rendered without `toYaml`. One site is the whole fix.

## Verification

`helm template` with a service carrying `args: ["setarch","-R","/bin/echo","hello"]`:

| | parsed `args` | length |
|---|---|---|
| before | `["setarch -R /bin/echo hello"]` | 1 |
| after | `["setarch","-R","/bin/echo","hello"]` | 4 |

- The new test asserts on the **parsed list**, not a substring. `"setarch" in str(args)` holds both before and after, so a substring assertion would pass either way and prove nothing. Confirmed the test discriminates by running it against the unmodified template (fails) and with the fix (passes), in that order.
- Rendering the **default** chart is byte-identical before and after, since the branch only runs when `args` is set.
- `pytest test/k8s_sandbox/helm_chart/`: 41 passed. `ruff check` and `ruff format --check` clean.
- Client-side `helm template` only (helm v3.16.3), no cluster needed, consistent with the rest of `test_helm_chart.py`.

## Two notes for the reviewer

1. **CHANGELOG placement.** I added a `## Unreleased` heading, following the pattern from 0.5.0 through 0.13.0 where contributors accumulate under `Unreleased` and the release commit renames it. `0.13.0` is already on PyPI and this tree is `0.14.0`, so the entry did not seem to belong under the `2026-08-12 0.13.0` heading. Happy to move it if you would rather keep one section.
2. **Possible trivial conflict.** #228 also touches `test_helm_chart.py`. I appended the new test immediately before the private helpers to keep the file's tests-then-helpers order and minimise overlap, but say the word if you would like it rebased after that lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,8 @@
 - The CoreDNS sidecar no longer serves its `ready` endpoint on port 8181, and refuses
   queries beyond 1000 concurrent.
 - Raise an error when a conflicting `max_pod_ops` setting would otherwise be ignored.
-- Fix a service's `args` rendering as one space-joined string rather than a list, so a
-  compose `command:` (which the converter maps to Helm's `args`) now reaches the container
-  as the argv the caller wrote instead of a single token. Using compose `entrypoint:` was
-  never affected, since that maps to Helm's `command`, which already used `toYaml`.
+- Fix a service's `args` (compose `command:`) reaching the container as a single
+  space-joined string instead of a list.
 
 ## 2026-08-12 0.13.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 - The CoreDNS sidecar no longer serves its `ready` endpoint on port 8181, and refuses
   queries beyond 1000 concurrent.
 - Raise an error when a conflicting `max_pod_ops` setting would otherwise be ignored.
+- Fix a service's `args` rendering as one space-joined string rather than a list, so a
+  compose `command:` (which the converter maps to Helm's `args`) now reaches the container
+  as the argv the caller wrote instead of a single token. Using compose `entrypoint:` was
+  never affected, since that maps to Helm's `command`, which already used `toYaml`.
 
 ## 2026-08-12 0.13.0
 
--- a/src/k8s_sandbox/resources/helm/agent-env/templates/services.yaml
+++ b/src/k8s_sandbox/resources/helm/agent-env/templates/services.yaml
@@ -92,7 +92,8 @@ spec:
           {{- toYaml $service.command | nindent 10 }}
         {{- end }}
         {{- if $service.args }}
-        args: {{ $service.args }}
+        args:
+          {{- toYaml $service.args | nindent 10 }}
         {{- end }}
         {{- if $service.workingDir }}
         workingDir: {{ $service.workingDir }}
--- a/test/k8s_sandbox/helm_chart/resources/service-args-values.yaml
+++ b/test/k8s_sandbox/helm_chart/resources/service-args-values.yaml
@@ -1,0 +1,5 @@
+services:
+  default:
+    image: "ubuntu"
+    command: ["/bin/sh", "-c"]
+    args: ["setarch", "-R", "/bin/echo", "hello"]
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+- Fix a service's `args` rendering as one space-joined string rather than a list, so a
+  compose `command:` (which the converter maps to Helm's `args`) now reaches the container
+  as the argv the caller wrote instead of a single token. Using compose `entrypoint:` was
+  never affected, since that maps to Helm's `command`, which already used `toYaml`.
```
art-dsit: Can you please shorten this a bit (end-users have to read this changelog and we want to make it easy for them to skip irrelevancies)

art-dsit (approved):

ppcvote:
Rebased onto `main` to clear the conflict, which was in `CHANGELOG.md` only. My entry now sits at the end of the `## Unreleased` list that #240 restored, and the duplicate heading this branch was carrying is gone. Code and tests are unchanged.

Thanks for #240 by the way, that is a better fix than the note I left here: restoring the released section and documenting the failure mode in CONTRIBUTING.md addresses the cause rather than just this instance.

Re-verified after the rebase rather than assuming it carried: `pytest test/k8s_sandbox/helm_chart/` is 41 passed, and the new test still discriminates, failing against `main`'s template and passing with the fix.

CHANGELOG.md
art-dsit: Shortened in d4547be (pushed to the branch directly).

PR #241, opened 2026-08-13
Add INSPECT_POD_RESTART_CHECK to skip the pre-op pod read in file operations

revmischa:
## What

Adds an opt-out for the `read_namespaced_pod` call that runs before every `read_file()` / `write_file()`:

```
INSPECT_POD_RESTART_CHECK=false
```

Defaults to enabled, so nothing changes unless a deployment sets it.

## Why

At high concurrency (200+ concurrent pod operations) that per-operation read becomes a meaningful load source on the Kubernetes API server — one extra API call for every file read and write, across every sandbox in an eval set.

## Why `exec()` is deliberately not gated

`exec()` keeps its check unconditionally, and that asymmetry is the point rather than an oversight.

The pre-op read is how we learn a sandbox pod was replaced underneath us. An agent `exec`s constantly, so detection-on-exec is both cheap in relative terms and the high-signal path — it is where a replaced pod is noticed first. File operations are different: an API error against a replaced pod is self-revealing without the pre-check, so the check earns less there while costing the same.

So the knob trades away the redundant half of the detection and keeps the half that does the work.

## Testing

`test/k8s_sandbox/pod/test_check_for_pod_restart.py` gains coverage for both settings across `read_file`, `write_file` and `exec`, asserting `exec` still checks when the variable is `false`.

```
9 passed
```

## Provenance

Originally written by @rasmusfaber for METR's fork, where it has been running in production since May. Commit authorship is preserved. We are upstreaming it rather than carrying a private divergence — it is currently the only commit in our fork that is not upstream, and a configurable knob defaulting to today's behaviour seemed more useful shared than held.

### DIFF
--- a/src/k8s_sandbox/_pod/pod.py
+++ b/src/k8s_sandbox/_pod/pod.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import logging
+import os
 from pathlib import Path
 from typing import IO, Callable, Literal, TypeVar
 
@@ -17,6 +18,15 @@ from k8s_sandbox._pod.write import WriteFileOperation
 T = TypeVar("T")
 
 logger = logging.getLogger(__name__)
+
+
+def _file_op_restart_check_enabled() -> bool:
+    # Gates the pre-op `read_namespaced_pod` call inside read_file / write_file
+    # (only — exec is unaffected and always checks). At high concurrency
+    # (200+ ops) the per-op reads can overwhelm the K8s API server; exec's
+    # check is the high-signal way we learn a sandbox pod was replaced and
+    # stays unconditional, while file-op API errors are self-revealing.
+    return os.environ.get("INSPECT_POD_RESTART_CHECK", "true").lower() != "false"
 
 
 class Pod:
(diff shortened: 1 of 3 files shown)


src/k8s_sandbox/_pod/pod.py
```diff
@@ -19,6 +20,15 @@
 logger = logging.getLogger(__name__)
 
 
+def _file_op_restart_check_enabled() -> bool:
+    # Gates the pre-op `read_namespaced_pod` call inside read_file / write_file
+    # (only — exec is unaffected and always checks). At high concurrency
+    # (200+ ops) the per-op reads can overwhelm the K8s API server; exec's
+    # check is the high-signal way we learn a sandbox pod was replaced and
+    # stays unconditional, while file-op API errors are self-revealing.
```
dpolatajko-aisi, line 28: nit: i agree that relying on the `exec` check is an acceptable trade-off and since this is opt-in behaviour I don't think it's blocking, but "file-op API errors are self-revealing" seems to imply that a file-op on a restarted pod would throw an API error, which wouldn't always be true - if the pod successfully restarted before the next file-op, there could be silent success edge cases

dpolatajko-aisi (approved):
lgtm - with one little nit on the documentation comment

PR #242, opened 2026-08-15
fix: skip runuser when the container is already the requested user

revmischa:
## The problem

`exec(user=...)` always wraps the shell in `runuser -u <user> /bin/sh`. `runuser` calls
`setgroups(2)`, which requires `CAP_SETGID` — **even when the switch is a no-op** because
the container already runs as that user. In a container whose capabilities have been
dropped, every such exec fails with:

```
runuser: cannot set groups: Operation not permitted
```

inspect-ai injects its sandbox tools (`text_editor`, `bash_session`, skills, checkpoint
snapshots) by probing with `user="root"`, so a sandbox hardened with
`securityContext: {capabilities: {drop: [ALL]}}` hits this on every injection.

**How visible that is depends on the container's user**, and the original wording here
was too broad (thanks @art-dsit for the correction):

| container | provider on `main` | inspect-ai's injector |
| -- | -- | -- |
| root, `CAP_SETGID` dropped | `cannot set groups` is not matched by `_check_for_runuser_error`, so a **failed `ExecResult`** | falls back to an exec with no `user`, which succeeds because the default user is also root — injection completes, and the defect is masked |
| non-root | raises on `runuser: may not be used by non-root users` | the exception propagates as a fatal `SandboxInjectionError`; no fallback is reachable |

So on a root container the failure is silent rather than fatal: the tools land, but as
the default user rather than root, so the `chmod 700` that hides them from the agent is
skipped. Direct `exec(user="root")` callers still get a failed result either way.

## The change

Check the current user in the container rather than on the client, so the fast path
costs no extra round trip:

```sh
sh -c 'if [ "$(id -un)" = X ] || [ "$(id -u)" = X ]; then exec /bin/sh; fi;
       exec runuser -u X -- /bin/sh'
```

- `sh -c` takes its script from argv, so stdin reaches whichever shell is exec'd unread
  and the caller's protocol is unchanged.
- A missing or unusual `id` leaves the substitution empty and falls through to
  `runuser` — the previous behaviour.
- Both the name and the uid are compared, so `user="1000"` works.
- The user is quoted in all three positions it now appears in.
- `--` before the shell so a username beginning with a dash cannot be read as an option.

### Knock-on: where "runuser not installed" is detected

`runuser` is now exec'd by the shell rather than by the API server, so a missing binary
surfaces as a 127 on stderr rather than as `ExecutableNotFoundError`. That explanation
moves into `_check_for_runuser_error`, alongside a new one naming `CAP_SETGID` for the
case above. Net effect on messages: unchanged for a missing binary, better for a dropped
capability.

One small behaviour change worth calling out: if the container is already the requested
user, `exec(user=...)` now **succeeds** on an image without `runuser` installed, where it
previously raised. That seems like the right way round, but say if you'd rather it stayed
strict.

## Testing

- New unit tests in `test/k8s_sandbox/pod/test_execute.py` covering the command
  construction (skip path, fallback path, uid match, quoting of adversarial usernames)
  and each `runuser` stderr explanation.
- `test/k8s_sandbox/pod` — 65 passed.
- `ruff check` / `ruff format --check` clean.

⚠️ **Not exercised against a real cluster.** I don't have one to hand for the
Helm-backed integration tests, so the `sh -c` wrapper's stdin behaviour is reasoned
about rather than observed. Worth a run before merge — happy to iterate if it misbehaves.

Draft for that reason, and because you may prefer a narrower fix (just the `CAP_SETGID`
error message, leaving the wrapper unconditional). I'm glad to cut it down if so.

---

## Update (since the original description)

Three things changed in review; the sections above are otherwise as written.

**Only the applicable comparison is emitted.** The wrapper originally tested both
`id -un` and `id -u`. That is wrong for an account merely *named* `0`, which would match
uid 0 and run the command as the wrong user. A numeric `user` is now compared against
`id -u` and a name against `id -un`, never both. `user="1000"` still works.

**A container that cannot switch users warns instead of raising.** This is the
user-visible contract change in the PR, so it deserves its own line. Of the four
`runuser` diagnostics, two describe an *environment* that cannot perform the switch — a
non-root container, and one without `CAP_SETGID` — rather than a caller asking for
something that does not exist. Those now log a warning and return the failed
`ExecResult`. An unknown user and a missing `runuser` still raise.

The reason is that callers are entitled to handle the first two. inspect-ai probes with
`user="root"` when injecting its sandbox tools and falls back to the container's own user
if that fails, which is how a rootless sandbox is meant to work. Raising made the
fallback unreachable, so a non-root container could not run *any* task using
`text_editor`, `bash_session` or skills — the exception propagated as a fatal injection
error. `test_exec_user_when_not_running_as_root` moves onto the new contract, and
`limitations.md` now states which failures raise and which don't. Happy to revert this
half if you'd rather keep both strict; it is separable from the wrapper fix.

**Validated on a real cluster**, which the original description flagged as missing — see
the comment below for what was run.

### Versions behind the downstream report

- **inspect-ai `0.3.260.dev5+g4a59f8538`.** Its injector (`_sandbox_tools_utils/sandbox.py`)
  treats a failed `ExecResult` as recoverable and falls back to an exec with no `user`,
  but turns any *exception* into a fatal `SandboxInjectionError`. That fallback has been
  in place since `0.3.240` (`643945dc7`, 2026-06-09), so it predates the report.
- **Image:** a root-by-default Debian-based task image; the sandbox ran as root with
  `capabilities: {drop: [ALL]}` and `seccompProfile: RuntimeDefault` under gVisor.
- The reported symptom was `runuser: cannot set groups: Operation not permitted` during
  tool injection on that root image — which, per the table above, is the masked case
  rather than the fatal one. The original description overstated it as fatal for root
  containers too; corrected above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,9 @@
 - `exec(user=...)` no longer wraps the shell in `runuser` when the container is already
   running as that user. `runuser` calls `setgroups(2)`, which needs `CAP_SETGID` even
   for a root -> root switch, so the unconditional wrapper made every `exec(user=...)`
-  fail in a container whose capabilities had been dropped. A missing `runuser`, a
-  missing `CAP_SETGID` and an unknown user are now each reported with an explanation
+  fail in a container whose capabilities had been dropped. On that path the process
+  environment is the container's rather than one `runuser` has reset (`HOME`, `USER`,
+  supplementary groups). A dropped `CAP_SETGID` is now reported with an explanation
   rather than a raw `runuser` error.
 
 ## 2026-08-12 0.13.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,10 @@
   for a root -> root switch, so the unconditional wrapper made every `exec(user=...)`
   fail in a container whose capabilities had been dropped. On that path the process
   environment is the container's rather than one `runuser` has reset (`HOME`, `USER`,
-  supplementary groups). A dropped `CAP_SETGID` is now reported with an explanation
-  rather than a raw `runuser` error.
+  supplementary groups). A dropped `CAP_SETGID` and a non-root container are now
+  logged as warnings and returned as a failed `ExecResult` rather than raised, so a
+  caller that probes with a user and falls back (as inspect-ai does when injecting
+  its sandbox tools) can do so. An unknown user and a missing `runuser` still raise.
 
 ## 2026-08-12 0.13.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,10 +19,11 @@
   for a root -> root switch, so the unconditional wrapper made every `exec(user=...)`
   fail in a container whose capabilities had been dropped. On that path the process
   environment is the container's rather than one `runuser` has reset (`HOME`, `USER`,
-  supplementary groups). A dropped `CAP_SETGID` and a non-root container are now
-  logged as warnings and returned as a failed `ExecResult` rather than raised, so a
-  caller that probes with a user and falls back (as inspect-ai does when injecting
-  its sandbox tools) can do so. An unknown user and a missing `runuser` still raise.
+  supplementary groups). A container that cannot switch users at all -- non-root, no
+  `CAP_SETGID`, or no `runuser` installed -- is now logged as a warning and returned as
+  a failed `ExecResult` rather than raised, so a caller that probes with a user and
+  falls back (as inspect-ai does when injecting its sandbox tools) can do so. Naming a
+  user that does not exist still raises.
 
 ## 2026-08-12 0.13.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@
 - Raise an error when a conflicting `max_pod_ops` setting would otherwise be ignored.
 - Fix a service's `args` (compose `command:`) reaching the container as a single
   space-joined string instead of a list.
+- `exec(user=...)` no longer wraps the shell in `runuser` when the container is already
+  running as that user. `runuser` calls `setgroups(2)`, which needs `CAP_SETGID` even
+  for a root -> root switch, so the unconditional wrapper made every `exec(user=...)`
+  fail in a container whose capabilities had been dropped. A missing `runuser`, a
+  missing `CAP_SETGID` and an unknown user are now each reported with an explanation
+  rather than a raw `runuser` error.
 
 ## 2026-08-12 0.13.0
 
--- a/docs/docs/design/limitations.md
+++ b/docs/docs/design/limitations.md
@@ -178,8 +178,14 @@ started with is **not recommended**. Generally, specifying users in tool definit
 result in undesirable coupling between your tools and sandbox.
 
 That said, if you need to run commands as different users, the `user` parameter to
-`exec()` is supported. However, you must run the container as root and ensure that
-`runuser` is installed in the container.
+`exec()` is supported. To switch to a *different* user, you must run the container as
+root, ensure that `runuser` is installed in the container, and keep `CAP_SETGID` — the
+switch goes through `runuser`, which calls `setgroups(2)`. A container with
+`capabilities: {drop: [ALL]}` therefore cannot switch users.
+
+Naming the user the container already runs as needs none of those: that case skips
+`runuser` entirely, so it works on a capability-dropped or non-root container, and on an
+image with no `runuser` installed.
 
 ## Images are not automatically built, tagged or pushed
 
--- a/docs/docs/design/limitations.md
+++ b/docs/docs/design/limitations.md
@@ -187,11 +187,12 @@ Naming the user the container already runs as needs none of those: that case ski
 `runuser` entirely, so it works on a capability-dropped or non-root container, and on an
 image with no `runuser` installed.
 
-When the switch cannot be made, the two causes are reported differently. A container
-that is not root, or that has had `CAP_SETGID` dropped, is a property of the environment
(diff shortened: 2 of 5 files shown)


revmischa:
## Now validated against a real cluster

The PR description flagged that the `sh -c` wrapper's stdin behaviour was reasoned about rather
than observed. It has now been exercised on an EKS cluster (Cilium, gVisor RuntimeClass), via a
downstream build pinning these commits.

Setup: sandbox pods with `securityContext: {capabilities: {drop: [ALL]}}`, i.e. no `CAP_SETGID`,
container running as root.

- Before: inspect-ai's sandbox tool injection (which execs `user="root"`) failed with
  `runuser: cannot set groups: Operation not permitted`, so every task using `text_editor` /
  `bash_session` died.
- After: the same task reports `File created successfully at: /tmp/injected.txt` with `error: null`,
  and the sandbox still shows `CapEff: 0000000000000000`.

Plain `exec()` with no `user` (the `["/bin/sh"]` path, unchanged) was run alongside it as a control
and behaves identically to before, returning correct stdout and exit codes.

So on a live cluster: stdin reaches whichever shell is exec'd, unread; the sentinel and returncode
handling is unaffected by the extra `sh -c` layer; and a capability-dropped root container can now
`exec(user="root")`.

art-dsit:
Hi, happy to merge this as-is. Sol got confused while trying to reproduce, so if you fancy wading through the following slop (or pointing your agent at it) that might improve things

BEGIN AI SLOP

I reproduced the root case on EKS, but the distinction between the provider failure and Inspect's fallback is important.

With both `ubuntu:24.04` (util-linux 2.39.3) and `debian:12` (util-linux 2.38.1), running as root with `capabilities.drop: [ALL]`:

```text
                         main                         this PR
plain exec capabilities  CapEff/CapPrm/CapBnd = 0    CapEff/CapPrm/CapBnd = 0
direct runuser root→root  rc 1: cannot set groups    rc 1: cannot set groups
exec(user="root")         failed ExecResult, rc 1     success, rc 0
```

That validates the fast path: `runuser` attempts the unnecessary group setup without `CAP_SETGID`, while this PR skips it and successfully executes as the already-current user.

However, `text_editor` injection succeeded for me on `main`. The existing provider does not recognise `cannot set groups` as one of the `runuser` errors it raises, so it returns a failed `ExecResult`; current Inspect then falls back to an exec without `user`, which succeeds because the container's default user is also root. Successful tool injection therefore does not mean the `user="root"` probe succeeded.

Could we update **The problem** and the validation comment to distinguish these outcomes? In particular, “every task using those tools failed, on any image, root or not” is broader than the evidence and does not hold with the current Inspect fallback in the versions I tested. I suggest something like:

> In a root container with `CAP_SETGID` dropped, `exec(user="root")` unnecessarily invokes `runuser`. On Ubuntu 24.04/util-linux 2.39.3 and Debian 12/util-linux 2.38.1, this returns exit code 1 with `runuser: cannot set groups: Operation not permitted`, even though no identity change is required.
>
> Inspect's sandbox-tool injector probes with `user="root"` and may recover by falling back to the container's default user. This can mask the provider defect during tool injection, but direct `exec(user="root")` callers still receive a failed result. Non-root containers are a separate case: the provider currently raises on `runuser: may not be used by non-root users`, preventing Inspect's fallback; this PR changes that failure to a warning and failed result.

It would also help to name the inspect-ai commit/version, image and `runuser --version` used by the downstream validation. That would make it clear whether its fatal tool-injection result predates or otherwise differs from current Inspect's fallback behaviour.

One code point before merge: a missing `runuser` is still raised, while the non-root and missing-`CAP_SETGID` cases now warn and return the failed `ExecResult`. The rationale says callers should be allowed to handle environmental failures, but absence of `runuser` is also an environmental failure for which Inspect's fallback helps. A rootless BusyBox/Alpine-style image where the root probe targets a different user would therefore still raise before Inspect can retry as the default user.

Could the missing-`runuser` branch also warn and return the failed result, with an integration test covering that fallback? If preserving the exception is intentional, the description should explain why this environment failure has a different contract from the other two.

END AI SLOP

art-dsit (approved):

revmischa:
Thanks — both points were right, and the second one changed the code.

**On the framing.** You're correct and I've rewritten *The problem*. I'd conflated "the provider returns a failure" with "the eval dies", and they're only the same thing in the non-root case. Confirmed against inspect-ai `0.3.260.dev5+g4a59f8538`: `_inject_container_tools_code` treats a failed `ExecResult` as recoverable and retries with no `user`, but turns any *exception* into a fatal `SandboxInjectionError`. Since `main` doesn't match `cannot set groups`, a root container gets the recoverable path — so injection completes and the defect is masked, exactly as you saw. That fallback has been in inspect-ai since `0.3.240` (`643945dc7`, 2026-06-09), so it predates our report rather than postdating it; the downstream image was root-by-default Debian-based, under gVisor with `drop: [ALL]` and `RuntimeDefault`. The description now carries that table and those versions.

**On the missing-`runuser` contract.** You've convinced me — it now warns and returns the failed result like the other two. The line I had in mind was "fail on image defects, warn on deployment posture", but that distinguishes who caused the problem rather than whether the caller can do anything about it, and the caller's options are identical in all three. Keeping the raise just meant one recoverable failure killed the run while two others didn't, with nothing signalling which you'd hit.

Only an unknown user still raises, since no fallback makes an absent account work.

`test_exec_user_when_runuser_not_installed` now asserts the recovery rather than the exception: the `exec` with a user fails and warns, and a follow-up `exec` without one succeeds on the same busybox sandbox. Also rebased onto `main` — the conflict was CHANGELOG-only.

revmischa:
@art-dsit all set?

PR #245, opened 2026-08-17
fix(helm): make network isolation subtractive so layered policies are not shadowed

rasmusfaber:
## Problem

A service with `network_mode: none` gets `networkIsolated: true`, which renders `svc-<n>-isolate` with a
blanket `ingressDeny: [{fromEntities: [all]}]`. Cilium resolves deny before allow with no exception
mechanism, so **any** ingress policy layered on top of this chart by a deployment is silently shadowed.

METR hit this attaching a policy that allows TCP 2222 to the `default` service so a human evaluator can SSH
into a sandbox and perform the task by hand. On a task that isolates `default`, the sandbox listens on 2222
and drops every packet, while the tooling still prints a connect command. Advertised, and silently
impossible.

## Why not just carve a port out of the deny

That was tried first and cannot work. Two reasons:

1. This chart's per-service ingress allow emits `toPorts` only under `{{- if $svc.ports }}`, so for a
   service with no declared ports — the normal case for `network_mode: none` — it is an **L3 allow covering
   every IP protocol**. A port-scoped deny cannot cancel it.
2. Cilium's `toPorts` cannot express a deny for protocols like GRE, ESP or AH, so a port-scoped deny can
   never reach parity with the blanket one.

## Change

Make isolation subtractive: stop granting the isolated service anything and let the chart's existing
default-deny do the work.

- Skip the per-service `svc-<n>-ingress` allow entirely when `$svc.networkIsolated` is set. Previously it
  was rendered and then cancelled by the blanket deny. (The `.Values.networks` branch already behaves this
  way, since a service cannot set both `network_mode: none` and `networks`.)
- Drop `ingressDeny` from `svc-<n>-isolate`, keeping `egressDeny` unchanged.

`sandbox-default-deny-ingress` already selects every pod in the release with `ingress: [- {}]`, so with no
allow selecting it, an isolated service is denied everything — including the protocols a `toPorts` deny
cannot express.

## Behaviour change

For a chart used on its own, observable behaviour is unchanged: the service was fully denied before and is
fully denied now. What changes is robustness. Isolation now rests on "no allow selects this service" rather
than on an unconditional deny, so a policy layered on top **will** take effect rather than being shadowed.
That is the point of the change, but it is a real semantic difference and is called out in the CHANGELOG.

## Verification

Beyond the chart tests, the resulting policy shape was verified on a live cluster running Cilium v1.18.4,
with two pods from a real image and the sandbox running a real SSH daemon on 2222 and 8000:

| | result |
|---|---|
| external → 2222 | open, full SSH key exchange completes |
| external → 8000 | blocked |
| sibling pod → 2222 | open (only because a deployment-supplied allow permits it) |
| sibling pod → 8000 | blocked |
| sibling pod → ICMP echo | blocked, with no ICMP rule present at all |
| egress from the isolated pod | dead (DNS and HTTPS both fail) |

Reply traffic on an allowed inbound connection flows despite the retained blanket `egressDeny`, because
Cilium is conntrack-based.

`test_network_isolated_service` is updated to assert the new shape, and additionally that no ingress allow
is rendered for the isolated service while a normal service still gets one.

`uv run pytest test/k8s_sandbox/helm_chart test/k8s_sandbox/compose -q` → 167 passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- `network_mode: none` isolation is now enforced by omitting any ingress allow for the
+  service rather than an unconditional ingress deny. Observable behaviour is unchanged
+  for a chart used on its own, but a network policy layered on top of this chart (e.g.
+  to allow a specific port) now takes effect instead of being silently shadowed.
 - **BREAKING CHANGE**: The CoreDNS sidecar now runs as UID/GID 65532 on a read-only root
   filesystem with only `NET_BIND_SERVICE`. A custom `corednsImage` must run under that
   context; set the new `corednsSecurityContext` if it cannot. The default image moves
--- a/src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
+++ b/src/k8s_sandbox/resources/helm/agent-env/templates/network-policy.yaml
@@ -138,7 +138,7 @@ spec:
     - {}
 {{- if .Values.networks }}
   {{- range $svcName, $svc := .Values.services }}
-    {{- if $svc.networks }}
+    {{- if and $svc.networks (not $svc.networkIsolated) }}
       {{- range $net := $svc.networks }}
 ---
 apiVersion: cilium.io/v2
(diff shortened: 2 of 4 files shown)


dpolatajko-aisi (approved):
lgtm 🚀

PR #249, opened 2026-08-31
Start sandboxes as soon as the cluster can schedule them

rasmusfaber:
## Summary

A sandbox that cannot be scheduled yet holds one of the eight install slots for as long
as it waits, so sandboxes the cluster *could* start immediately queue behind ones it
can't — and because their pods are never created, the autoscaler never sees the demand
it would satisfy. Measured on a real cluster (12 releases, 4 permanently unschedulable,
4 slots): the first runnable sandbox became ready at 327.6s, of which ~180s was purely
queueing and the rest was node provisioning that was always irreducible. This releases
the slot once Helm has created the release and waits for readiness off it — 156.0s, with
install-slot-seconds across the run dropping from 1,368 to 88.8.

Reimplementing that wait exposed a second bug worth fixing in the same PR. Readiness
compared a *count* of ready pods against a count derived from the manifest, so any pod
carrying the release label could satisfy it — a DaemonSet's, or a Helm hook's. Since
Inspect takes the first entry of the returned sandbox dict as the default sandbox, that
hands the agent whichever container happened to come up and scores the sample with no
error anywhere. Readiness now waits for the sandboxes the chart *names*, through the
same `inspect/service` label that handover keys on.

## Evidence

Against a real EKS cluster, on the branch:

```
$ INSPECT_K8S_DEFAULT_NAMESPACE=... uv run pytest test/k8s_sandbox/inspect_integration/ \
    test/k8s_sandbox/test_helm.py -q -m req_k8s -o "env=INSPECT_HELM_TIMEOUT=420"
29 passed, 106 deselected in 1429.51s (0:23:49)
```

That includes the four `custom_chart` tests. The chart there renders a bare `kind: Pod`,
which is the shape a StatefulSet-based readiness check silently skips — it is the
regression test for the bug the second commit fixes.

`uv run pytest -q -m "not req_k8s"` → `452 passed, 249 deselected`.

Negative control, reverting each source change and re-running the non-cluster suite:

```
permit held across the wait          → test_install_permit_is_released...  fails
empty declaration = nothing to wait  → 1 failed
traversal reads controller metadata  → 1 failed
readiness drops the per-sandbox req  → 2 failed
handover re-lists instead of reusing → 1 failed
Succeeded pod stands in for sandbox  → 1 failed
allow_stale / read timeout / label selector / deadline placement → each fails
```

Two claims about Helm rather than about this code, measured rather than derived:

```
$ helm install --timeout=5s          <blackholed apiserver>   → fails after 30.04s
$ helm install --timeout=5s --wait   <blackholed apiserver>   → fails after 30.12s
```

`--timeout` does not bound getting the release created, with or without `--wait`, so the
last commit corrects docs that said it did. And `helm install` rejects a numeric service
name outright (`cannot unmarshal number into ObjectMeta.metadata.labels`), so the
label-typing hazard that shape suggests is unreachable.

## Notes for reviewers

Deliberately not here: an *explicit* declaration of expected sandbox names. Inferring
them from labels cannot catch an author who omits the label entirely, and only a declared
list can — but that is a public contract change and wants its own discussion.

## Agent review

Written by Claude Opus 5 (Claude Code) working interactively with me. Reviewed by
five Claude Opus 5 subagents (bug-finder, test-reviewer, type-checker, code-simplifier,
code-structure) and by GPT-5.6 via Codex, over two rounds; the readiness design was
additionally wargamed by three Claude Opus 5 architects and Codex before being rewritten.

What they found that mattered: the count-vs-set bug above; that a set of names alone is
also wrong because it loses cardinality; that treating "no sandbox declared" as "nothing
to wait for" repeats the same class of bug; that the diagnostics reads were unbounded on
the one path reached when the API is unresponsive; and that nine of fourteen mutations
survived the suite at one point. Three earlier readiness designs of mine were rejected
during this process, including one I had already pushed.


src/k8s_sandbox/_helm.py
```diff
@@ -241,136 +283,81 @@ async def uninstall(self, quiet: bool) -> None:
...
-                    self._namespace,
-                    self._context_name,
-                    default_container_name,
-                    pod.uid,
-                    pod.restart_count_for(default_container_name),
-                    self.restarted_container_behavior,
-                )
+        for pod in _sandbox_pods(pods):
+            default_container_name = pod.container_names[0]
+            sandboxes[pod.labels[_SERVICE_LABEL]] = Pod(
```
dpolatajko-aisi, line 303: there is a possible edge case here where there are multiple pods with the same service label in different states (e.g. a healthy one and a completed one) and the pod which is chosen is dependent on the list order, which can result in `exec` failing if the map points to a completed pod. i reproduced this failure, and although i think its unlikely to be an issue in most cases, i think filtering out terminal pods here would be safe and easy

dpolatajko-aisi (commented):
this mostly seems good to me, and i reproduced the performance boost. thanks for the docs as well. i have one question about an edge case before i ✅

dpolatajko-aisi (approved):
thanks for the update

PR #250, opened 2026-08-31
Stop pods piling up when a sample's sandbox fails to start

rasmusfaber:
## Summary

When a sandbox fails to come up — `helm install` timing out, an image that won't pull —
`sample_init()` raises, and Inspect does not call `sample_cleanup()` in that case, because
`sample_init` is awaited outside the `try` that guards it. Nothing else removes the
release until the whole `eval()` ends, and `eval_set` retries a task up to ten times
*within the same process* by default, so a task that keeps failing accumulates a fresh set
of sandbox pods on every attempt while none of the previous sets are reclaimed. We hit
this on a production cluster: pods grew without converging until the cluster ran out of
capacity.

This uninstalls the release when `sample_init()` fails, so a retried sample leaves nothing
behind. A failed uninstall is suppressed rather than raised — it would otherwise replace
the install error, which is the one carrying the pod diagnostics — and the release stays
tracked so `uninstall_all()` retries and reports it at the end of the eval.

## Relationship to #248

#248 diagnoses the same bug, and I would be glad to see either land. The difference is
placement: in `Release.install()` a failing uninstall replaces the original error, the
manager goes on tracking a release that no longer exists (so it is uninstalled a second
time at shutdown, and listed under `--no-sandbox-cleanup` as still present), and the same
leak one line further on — `get_sandbox_pods()` raising after a successful install — is not
covered. `sample_init()` is the only production caller of both, so one guard there covers
both paths and keeps the manager's tracking honest. It also needs no changes to existing
tests, where #248 has to relax `test_helm_install_error`'s `spy.call_count` from 1 to 2 —
that relaxation is the double-uninstall showing up.

## Evidence

Reproduced without a cluster first: three retried `sample_init()` calls produce three
`helm install`s, **zero** `helm uninstall`s, and three leaked releases.

Then end to end on a real EKS cluster, an eval-set with 2 epochs and `retry_attempts=3`,
pods failing on image pull:

```
                    peak pods   peak releases   at shutdown
main                  8  (2→4→6→8, monotonic)   6    all 8 uninstalled at once
with this change      2  (flat, dips to 0)      0–1  nothing left
```

Test runs on the branch:

```
$ uv run pytest -q -m "not req_k8s"
417 passed, 248 deselected in 11.38s

$ INSPECT_K8S_DEFAULT_NAMESPACE=... uv run pytest \
    test/k8s_sandbox/inspect_integration/test_cleanup.py test/k8s_sandbox/test_helm.py \
    -q -m req_k8s -o "env=INSPECT_HELM_TIMEOUT=420"
12 passed, 74 deselected in 332.20s (0:05:32)
```

Negative control — reverting the source change and keeping the tests:

```
FAILED test_failed_install_uninstalls_the_release
FAILED test_failed_uninstall_does_not_mask_the_install_error
FAILED test_release_is_uninstalled_when_its_pods_cannot_be_read
3 failed, 8 passed
```

The remaining `req_k8s` tests (pod exec, read/write, networking) were not run; this change
touches none of those paths.

## Notes for reviewers

Cancellation is deliberately untouched. `Release.install()`'s existing
`except asyncio.CancelledError` handler does not actually run under anyio — a cancelled
scope re-delivers at every checkpoint, so the uninstall dies at the semaphore before Helm
is spawned — but those releases are still reclaimed by the shielded `uninstall_all()` at
shutdown, and shielding the per-sample uninstall would make Ctrl-C block behind N
`helm uninstall --wait` calls. Worth knowing that `test_cancelling_install_uninstalls`
passes only because raw `asyncio.Task.cancel()` delivers once; it is green while the
handler is inert in production.

## Agent review

Written by Claude Opus 5 (Claude Code) working interactively with me, and reviewed by
Claude Opus 5 subagents and by GPT-5.6 via Codex. Their findings that changed the patch:
that a failed uninstall would otherwise mask the diagnostics-bearing install error; that
the manager, not `Release`, owns the tracking that has to stay truthful; and that the
`get_sandbox_pods()` path leaks identically and needed the same guard.


src/k8s_sandbox/_sandbox_environment.py
```diff
@@ -225,8 +225,21 @@ def reorder_default_first(
...
+        except Exception:
+            # Inspect does not call sample_cleanup() when sample_init() raises, and
+            # uninstall_all() does not run until the whole eval ends, so nothing else
+            # removes what Helm created. Without this, each retried sample leaves a
+            # full set of sandbox pods behind.
+            # A failed uninstall must not replace the original error (which carries the
+            # pod diagnostics), and leaves the release tracked so that uninstall_all()
+            # retries it and reports it.
+            with suppress(Exception):
+                await manager.uninstall(release, quiet=True)
```
dpolatajko-aisi, line 241: could we please add handling for `--no-sandbox-cleanup` here? i think it can still be useful for debugging to honour that flag for init failures

rasmusfaber: This is unfortunately not simple to do. It will need a change to Inspect to somehow give access to the flag to `sample_init()`. Do you feel this is a blocker or a nice-to-have?

dpolatajko-aisi: ah yeah that is a good point. i think it's probably nice to have

dpolatajko-aisi (approved):

PR #252, opened 2026-09-03
Stop sandboxes from seeing their own Kubernetes namespace

rasmusfaber:
## Summary

The kubelet writes a pod's FQDN — `<pod>.<subdomain>.<namespace>.svc.cluster.local` — into `/etc/hosts`, so anything running in a sandbox can read the namespace it is in. Deployments that name the namespace after the eval hand that straight to the agent under evaluation: we found this after an agent's reasoning trace quoted its own namespace and correctly identified the model and benchmark it was being tested on.

A pod only gets that FQDN line when it has a subdomain, and a StatefulSet's pods take their subdomain from `spec.serviceName`. This chart set it to `<service>-service`, a Service the chart never creates — so it bought nothing and leaked the namespace. Emptying it makes the kubelet write a plain `<pod IP> <pod name>`, and `/etc/hosts` stays the kubelet's own file: still writable, still mapping the pod's real address.

## Notes for reviewers

- **Per-pod DNS (`<pod>.<service>.<namespace>.svc.cluster.local`) no longer resolves.** Nothing in this chart used it — the coredns sidecar rewrites short service names to the Services rendered alongside the StatefulSets — but a deployment supplying its own governing Service to address pods by name would need the pod IP instead.
- **No opt-out value**, deliberately: nothing observable breaks, and a chart-global switch is the wrong granularity for this (one auxiliary service needing the old behaviour would un-mask the agent's service too). If you'd rather users could restore it, I'd suggest `governingServiceName` (string, default `""`) over a boolean — an absent value then defaults to the safe behaviour, which matters for `helm upgrade --reuse-values` on existing releases.

## Test plan

Verified on a real cluster (EKS 1.35, gVisor runtime) with a two-service sandbox:

- `/etc/hosts` is the kubelet's own file and reads `10.112.206.114  agent-env-<rel>-default-0` — no FQDN. `hostname -f`, `hostname -A`, `hostname -d` and `socket.getfqdn()` are all namespace-free.
- Grepping the namespace string across the sandbox filesystem (24,569 files) finds nothing. Before the change the same sweep matched `/etc/hosts`, and only `/etc/hosts`.
- `/etc/hosts` is still writable: append and truncate both succeed and take effect (`getent hosts` resolves an appended entry).
- Service-to-service is unaffected: short service name, the legacy `$AGENT_ENV-<service>` name, raw TCP to a sibling pod IP, and `ping` all work.
- `serviceName: ""` is accepted by the apiserver on 1.31, 1.32 and 1.35 (validation has never required the field; the check added in 1.33 is guarded on it being non-empty). Note the OpenAPI schema still marks it required, so a strict third-party validator could object.

The new `test_dns.py` assertion is `req_k8s` and hasn't run in my environment — it needs a gvisor-capable cluster, so it is unexercised until your CI picks it up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- `network_mode: none` isolation is now enforced by omitting any ingress allow for the
+  service rather than an unconditional ingress deny. Observable behaviour is unchanged
+  for a chart used on its own, but a network policy layered on top of this chart (e.g.
+  to allow a specific port) now takes effect instead of being silently shadowed.
 - `INSPECT_POD_RESTART_CHECK=false` skips the pre-operation pod read inside
   `read_file()` / `write_file()`, for deployments where that per-op
   `read_namespaced_pod` call becomes a load problem on the Kubernetes API server at
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,9 @@
 - `exec(user=...)` no longer wraps the shell in `runuser` when the container is already
   running as that user. `runuser` calls `setgroups(2)`, which needs `CAP_SETGID` even
   for a root -> root switch, so the unconditional wrapper made every `exec(user=...)`
-  fail in a container whose capabilities had been dropped. A missing `runuser`, a
-  missing `CAP_SETGID` and an unknown user are now each reported with an explanation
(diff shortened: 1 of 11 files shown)


rasmusfaber:
Superseded: rebasing onto upstream main to drop unrelated fork patches from the diff.

PR #254, opened 2026-09-04
fix: honour compose `command:` on the default service

### LINKED ISSUE #237: compose `command:` on the `default` service never runs — chart-default `command` deep-merges over it
**SLOP ALERT**

## Summary

The chart's `values.yaml` defines `services.default.command: ["tail", "-f", "/dev/null"]`. The compose converter maps compose `command:` to Helm `args` and emits no `command` key when the compose file has no `entrypoint:`. Helm then deep-merges the chart default in, so the `default` service's entrypoint is always `tail -f /dev/null` and the user's compose `command:` is appended to it as extra arguments. The process the user asked for never runs.

This is distinct from #219/#236, which fix how `args` is *rendered*. After #236 the argv is at least the list the user wrote, but for the `default` service it is still appended to `tail -f /dev/null`. Only the service named `default` is affected (the only one with a chart-default `command`), and only when compose omits `entrypoint:`. The same compose file behaves correctly under Inspect's docker sandbox provider, where `command:` replaces the image CMD and the image's ENTRYPOINT still applies.

## Repro

```yaml
# compose.yaml
services:
  default:
    image: busybox:1.37
    command: ["sh", "-c", "touch /tmp/i-ran && tail -f /dev/null"]
```

Run any Inspect task against this sandbox on minikube (it works under `--sandbox docker`), and in the solver exec `ls /tmp/i-ran` and `tr '\0' ' ' < /proc/1/cmdline`. Observed on `main` (8a98e9f, pre-#236):

```
ls /tmp/i-ran: rc=1 ls: /tmp/i-ran: No such file or directory
pid1 argv: tail -f /dev/null sh -c touch /tmp/i-ran && tail -f /dev/null
```

The sandbox comes up looking healthy — `tail` warns about the unopenable extra arguments and keeps running — so the eval proceeds with a container that silently never ran the user's command.

With the #236 fix applied the same compose file instead fails loudly: the argv becomes `tail -f /dev/null sh -c 'touch ...'`, busybox `tail` exits non-zero on the `-c`, and the pod crashloops until `helm install --wait` times out:

```
pod_diagnostics: container 'default': terminated Error (exit code 1), restarted 4 time(s)
```

Which variant you get depends on whether any appended argument parses as a `tail` option, i.e. on the user's command and the image's `tail` implementation.

## Suggested fix

Have the converter emit an explicit `command: null` for a service that sets `command:` without `entrypoint:`. Helm treats a null in user values as "delete the default key", so the pod falls back to the image's ENTRYPOINT — matching docker compose semantics. (Needs verifying that the null survives the provider's values plumbing.) Removing the default from the chart's `values.yaml` instead would break the bare default chart, whose `python:3.12-bookworm` service relies on `tail -f /dev/null` to stay alive.

apurv-1:
Closes #237.

## The defect

The built-in chart sets `services.default.command: ["tail", "-f", "/dev/null"]`.
The compose converter maps compose `command:` to Helm `args` and emits no
`command` key when the compose file has no `entrypoint:`.

Helm then deep-merges the chart default in, so pid 1 becomes:

```
tail -f /dev/null sh -c 'touch /tmp/i-ran && tail -f /dev/null'
```

The user's command never runs. After #236 the argv is at least the list the
user wrote, but it is still appended to `tail`. Only the service named
`default` is affected (the only one with a chart-default `command`), and only
when compose omits `entrypoint:`. The same compose file is correct under
Inspect's docker sandbox.

## The fix

When a service sets compose `command:` without `entrypoint:`, the converter
now emits `command: null`. Helm treats that as "delete the default key", so
the pod falls back to the image ENTRYPOINT and the user's `args` become CMD.

That is the Helm-documented way to unset a default. Removing the default from
the chart's `values.yaml` instead would break the bare default chart, whose
`python:3.12-bookworm` service relies on `tail -f /dev/null` to stay alive.

Plumbing that had to change so the null survives through to Helm:

- `validate_no_null_values` now allows `services.<name>.command: null`
  (other nulls are still rejected)
- `values.schema.json` accepts `command: null`, matching `workingDir`

## Verification

`helm template` of the issue's compose file (via the converter):

| | rendered `command` | rendered `args` |
|---|---|---|
| before | `["tail", "-f", "/dev/null"]` | `["sh", "-c", "touch /tmp/i-ran && tail -f /dev/null"]` |
| after | omitted | `["sh", "-c", "touch /tmp/i-ran && tail -f /dev/null"]` |

- Converter tests cover command-only, entrypoint-only, both, YAML dump of
  `command: null`, chart-schema validation, and `validate_no_null_values`.
- Helm chart test asserts the default chart still renders
  `tail -f /dev/null`, and that `command: null` omits the entrypoint.
- `pytest -m "not req_k8s"` on the converter, helm, and chart tests:
  243 passed. `ruff check`, `ruff format --check`, and `mypy` on the
  changed sources are clean.
- Client-side `helm template` only (helm v3.21.4). No cluster.


dpolatajko-aisi (approved):
thanks for the contribution, looks good to me

PR #256, opened 2026-09-09
fix: wait for Cilium to apply a release's network policy

sjawhar:
## What happens today

Two separate things have to be true before a sandbox's first command is safe, and the
provider waits for neither.

**The pod is created before its policy.** Helm orders an install by kind and sorts kinds
it does not know — every custom resource, `CiliumNetworkPolicy` among them — after the
workloads. On this chart that is verifiable in the release itself:

```
$ helm get manifest <release> | grep ^kind:      # before this change
kind: ConfigMap
kind: ConfigMap
kind: StatefulSet
kind: CiliumNetworkPolicy
kind: CiliumNetworkPolicy
kind: CiliumNetworkPolicy
```

**`helm install --wait` returns on kubelet's signal.** A pod is Ready once its containers
are, which says nothing about Cilium having built that pod's endpoint and compiled its
policy. So the sandbox can be handed back while the endpoint is still being built, or
rebuilt because the policy arrived after it.

Either way the first command runs against a half-programmed egress policy. In a CI suite
which starts a sandbox and runs a fixed command immediately, with no LLM turn in between,
that shows up as:

```
AssertionError: assert 'EXAMPLE=OK' in
  'curl: (6) Could not resolve host: example.com\n\nEXAMPLE=BLOCKED\n'
```

— an *allow-listed* domain failing to resolve. Five occurrences in fourteen days across
five unrelated commits and three days, on identical package pins (so not a regression in
any of them); four of that shape, and one where a pod configured for unrestricted egress
had no network at all.

## The change

**The chart's network policies are installed before the pods they protect.** They carry
`helm.sh/hook: pre-install,pre-upgrade`, so Helm applies them in the hook phase, which
completes before the manifest phase that creates the StatefulSets:

```
$ helm get hooks <release> | grep ^kind:         # after this change
kind: CiliumNetworkPolicy
kind: CiliumNetworkPolicy
kind: CiliumNetworkPolicy
$ helm get manifest <release> | grep ^kind:
kind: ConfigMap
kind: ConfigMap
kind: StatefulSet
```

Helm does not remove hook resources with the release, so `uninstall()` deletes the
release's `CiliumNetworkPolicy` objects itself, after a successful `helm uninstall`
(never after a failed one, which would unpolice pods that are still running). It lists
them by the `app.kubernetes.io/instance` label the policies now carry and deletes each by
name, as Helm itself would: deleting a collection is authorized as the distinct
`deletecollection` verb, which an identity holding the verbs a chart installer needs
(`create`, `delete`, `get`, `list`, `patch`, `update`, `watch`) does not necessarily
have. Every failing delete is attempted and reported together, so one refusal does not
leave the rest behind. A release uninstalled by hand with `helm uninstall` keeps its
policies; `kubectl delete cnp -l app.kubernetes.io/instance=<release>` removes them.

**Permissions.** The wait reads `get` and `list` on `ciliumendpoints.cilium.io` in the
namespace it installs into; without them it is skipped with one warning per process.
The uninstall cleanup needs `list` and `delete` on `ciliumnetworkpolicies.cilium.io`.
`delete` was already required, since Helm deleted these objects itself while they were
ordinary release resources; `list` is new — an identity without it installs the chart
normally and then fails the uninstall with `Failed to list the release's network
policies`. Both are stated in `docs/getting-started/remote-cluster.md`.

**Installation then waits for Cilium to build the endpoints.** After a successful
`helm install`, every pod of the release (`app.kubernetes.io/instance=<release>`, the
selector `get_sandbox_pods` already uses) must have a `CiliumEndpoint` reporting
`status.state: ready` before the release is usable:

- An endpoint that does not exist yet is a state to wait through, not a pass.
- Endpoints are matched to pods by the owning pod's UID, not by name. Chart pods are
  StatefulSet members with stable names, so an endpoint left behind by a previous pod of
  the same name would otherwise end the wait immediately; matching on UID means a stale
  endpoint can only delay it.
- Host-networked pods share the node's network namespace and have no endpoint of their
  own, so they are not waited for.

`status.state` is the only signal available: the `cilium.io/v2` CRD's `status.policy`
declares just `ingress`/`egress` (an apiextensions/v1 CRD prunes anything else, so no
agent flag can add policy revisions to it), and `CiliumNetworkPolicy.status` carries no
per-node realization at all. Ordering the policies ahead of the pods is what makes that
signal sufficient rather than merely necessary: the endpoint's first build already has
the policy, so "built and not regenerating" is the whole race. What the hooks guarantee
is that the policy *objects* exist before the pod is created — not that every node's
agent has imported them before it builds that pod's endpoint. That import is the
remaining window, and it is milliseconds against the seconds a pod takes to be scheduled
and started.

The wait polls every 250 ms, backing off to 1 s after two seconds so a stalled cluster is
not also being listed four times a second per concurrent install, and gives up after 60 s.
What happens then depends on what it saw, because these are different faults:

- An endpoint exists but never became ready — the real fault — raises the new
  `CiliumPolicyNotRealizedError` (a `K8sError`, exported for callers to catch), naming the
  endpoints, their observed state and the elapsed time.
- No endpoint ever appeared for a pod is far more likely a node Cilium does not manage
  (Fargate, Windows, a cluster mid-rollout) than a stuck agent, so it warns naming those
  pods and continues, rather than turning a working sandbox launch into a hard failure.

Two cases are not failures at all, and are distinguished rather than collapsed into one
swallowed 404:

- **No `CiliumEndpoint` CRD** — the API 404s on the collection path itself. The cluster
  does not run Cilium; the wait is a no-op.
- **Not permitted to read them** — 401/403. The wait is skipped with a single warning per
  process naming the permission to grant, rather than breaking every sandbox on a cluster
  whose RBAC predates this version. `docs/getting-started/remote-cluster.md` states the
  requirement (`get`, `list` on `ciliumendpoints.cilium.io`) under `## Requirements`
  rather than under the built-in chart, because the wait runs for every chart.

Both reads — the release's pods and its endpoints — go through the same handling: a
throttled or briefly unavailable API server (including connection-level `urllib3` and
socket errors, which the kubernetes client does not convert into `ApiException`) is
retried inside the budget, and a read that never succeeds surfaces as a `K8sError` naming
the release rather than a bare client exception.

`PodSnapshot` gains `host_network` and the pod UID is now used; `_kubernetes_api` gains
`k8s_custom_objects_client` so custom-resource clients are built where the other clients
are.

## Verification

Unit tests (`test/k8s_sandbox/test_cilium.py`, faking the Kubernetes API at the same
boundary `test_snapshot.py` does): a ready endpoint returns on the first read; a missing
endpoint, a regenerating endpoint, a multi-pod release and an endpoint owned by a previous
pod of the same name each keep polling; the timeout raises naming the endpoint and its
state; a pod that never gets an endpoint warns once and continues; transient API and
connection errors are retried; reads that never succeed raise `K8sError` naming the
release; a 403 on either read warns once and skips; the CRD being absent is a silent
no-op; host-networked pods are not waited for; the policies are deleted one at a time
(never as a collection), one already gone is not an error, every failing delete is
attempted and named in one error, a policy the API returns without a name raises, and a
failed list is distinguishable from a failed delete. `test_helm.py` pins the sequence:
install waits, a failed install does not, uninstall deletes the policies, a failed
uninstall does not.

One `req_k8s` test pins the group, version, plural and status field against a real
cluster, because every other test in the file fakes the CRD — a typo in any of those four
strings would read as "this cluster does not run Cilium" and silently skip the wait.

```
$ uv run pytest test -m 'not req_k8s' -q
456 passed, 335 deselected in 29.26s
$ uv run ruff check . && uv run ruff format --check . && uv run mypy .
All checks passed!
86 files already formatted
Success: no issues found in 86 source files
```

Live, against a Cilium v1.18.4 EKS cluster, in namespaces created and deleted for the
purpose:

- Cluster tests: `test_cilium.py` and `test_helm.py -m req_k8s` → `10 passed in 36.10s`
  on this head; earlier on the same branch, with `test_network_policy.py`,
  `test_dns.py`, `test_networks.py` and both `test_network_policy_ports_*.py`,
  `19 passed in 97.05s` and `26 passed in 135.29s`.
- The `deletecollection` verb, tested rather than reasoned about: a ServiceAccount holding
  exactly the verbs a chart installer needs (`create`, `delete`, `get`, `list`, `patch`,
  `update`, `watch` on `ciliumnetworkpolicies`) — no `deletecollection` — running against
  a real release's policies gives
  `delete_collection_namespaced_custom_object` → `HTTP 403 Forbidden`, and the
  implementation's list-then-delete-by-name → ok, with `0` policies left.
- The new `list` requirement, same method: a ServiceAccount without
  `list` on `ciliumnetworkpolicies` installs the chart normally (`Install complete`) and
  then fails the cleanup with `Failed to list the release's network policies … (403)`,
  leaving three policies behind.
- A sandbox with `allowDomains: [example.com]` running
  `curl -sS https://example.com -o /dev/null -w '%{http_code}'` as its first command:
  `EXAMPLE=OK http='200' PYPI=BLOCKED`, with
  `Waited for Cilium to build the release's endpoints. {"release": "ukautsqp", "outcome": "realized", "elapsed_ms": "14", "endpoints": "agent-env-ukautsqp-default-0 (state ready)"}`,
  and nothing left in the namespace after uninstall (`after uninstall: 0 objects left`).
- Eight concurrent launches of the same sandbox before the ordering change: all returned
  `200`, and the wait cost 17–108 ms per sandbox with every endpoint already ready.
- The blocking path, live: deleting a release's pod makes Cilium build a fresh endpoint
  for the replacement, and the wait blocks on it —
  `Waiting for Cilium to build the release's endpoints. {"release": "gbla28nx", "pending": "agent-env-gbla28nx-default-0 (no CiliumEndpoint)"}`,
  then `BLOCKED for 572 ms`.
- The manifest/hooks output quoted above is from installs of this chart on that cluster,
  before and after the change.

One measurement changed the implementation: creating the Kubernetes client inside the
worker thread re-ran the kubeconfig's credential plugin per thread, costing ~0.9 s on
every concurrent sandbox start (847–1030 ms measured across eight). The clients are now
created on the caller's thread and used from the workers, as `get_sandbox_pods` does.

### DIFF
--- a/src/k8s_sandbox/_cilium.py
+++ b/src/k8s_sandbox/_cilium.py
@@ -1,0 +1,510 @@
+"""Wait for Cilium to program a release's pods before they are used.
+
+`helm install --wait` returns as soon as kubelet reports the release's pods
+Ready. Cilium builds each pod's endpoint and compiles its policy separately, so
+a sandbox handed back on kubelet's signal alone can run its first command
+against a half-programmed egress policy: an allow-listed domain that does not
+resolve, or a pod meant to have unrestricted egress with no network at all.
+
+Cilium publishes per-pod progress as a ``CiliumEndpoint`` (``cilium.io/v2``),
+named after the pod, owned by it, and carrying its labels. ``status.state`` is
+``ready`` only once that endpoint's datapath is built and not waiting to
+regenerate, and it is the only signal the API offers: the v2 CRD's
+``status.policy`` schema declares just ``ingress``/``egress`` (an
+apiextensions/v1 CRD prunes anything else), and ``CiliumNetworkPolicy.status``
+carries no per-node realization at all. So the wait covers the window in which
+the endpoint does not exist or is being (re)built. The chart's network policies
+are installed as pre-install hooks, which puts the policy objects in the API
+server before the pod is created; what remains unobservable is whether the
+node's agent had imported them by the time it built that endpoint.
+
+Reads use the raw API JSON rather than the kubernetes client's model
+deserialization, as ``k8s_sandbox._pod.snapshot`` does and for the same reason.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import functools
+import json
+import time
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, Sequence, cast
+
+import urllib3
+from kubernetes import client  # type: ignore
+from kubernetes.client.exceptions import ApiException  # type: ignore
+from urllib3 import HTTPResponse
+
+from k8s_sandbox._error import K8sError
+from k8s_sandbox._kubernetes_api import k8s_client, k8s_custom_objects_client
+from k8s_sandbox._logger import log_debug, log_warn
+from k8s_sandbox._pod.snapshot import PodSnapshot, list_pods
+
+CILIUM_GROUP = "cilium.io"
+CILIUM_VERSION = "v2"
+CILIUM_ENDPOINTS_PLURAL = "ciliumendpoints"
+CILIUM_NETWORK_POLICIES_PLURAL = "ciliumnetworkpolicies"
+CILIUM_ENDPOINTS_RESOURCE = f"{CILIUM_ENDPOINTS_PLURAL}.{CILIUM_GROUP}"
+
+READY_STATE = "ready"
+
+# Building an endpoint is a node-local control loop which completes in
+# milliseconds once the agent holds the policy, so poll fast enough that the
+# wait is invisible on the happy path.
+POLL_INTERVAL_SECONDS = 0.25
+# Past this the endpoint is not merely mid-build, so stop listing four times a
+# second per concurrent install against a cluster which is already struggling.
+SLOW_POLL_AFTER_SECONDS = 2.0
+SLOW_POLL_INTERVAL_SECONDS = 1.0
+# A minute of it means something is broken rather than busy — and failing is
+# better than handing back a sandbox whose egress rules are not in force.
+TIMEOUT_SECONDS = 60
+
+# The kubernetes client converts only SSLError into an ApiException, so a
+# connection-level failure reaches us as a urllib3 or socket error.
+_CONNECTION_ERRORS = (urllib3.exceptions.HTTPError, OSError)
+_READ_ERRORS = (ApiException, *_CONNECTION_ERRORS)
+
+
+class CiliumWaitOutcome(Enum):
+    """How a wait for Cilium to program the release's pods finished."""
+
+    REALIZED = "realized"
+    """Every endpoint the release's pods have was built and ready."""
+
+    CRD_ABSENT = "crd-absent"
+    """The cluster has no CiliumEndpoint CRD, so it does not run Cilium."""
+
+    NOT_PERMITTED = "not-permitted"
+    """The cluster runs Cilium but did not permit reading CiliumEndpoints."""
+
+    ENDPOINTS_ABSENT = "endpoints-absent"
+    """Pods which never got an endpoint: a node Cilium does not manage."""
+
+
+@dataclass(frozen=True)
+class CiliumEndpointStatus:
+    """A minimal view of one pod's CiliumEndpoint.
+
+    ``found`` is False for a pod whose endpoint Cilium has not published yet,
+    which is a state to wait through rather than a missing signal.
+    """
+
+    name: str
+    found: bool
+    state: str | None
+    pod_uid: str | None = None
+
+    @staticmethod
+    def absent(pod_name: str) -> CiliumEndpointStatus:
+        """The endpoint of a pod which Cilium has not published yet."""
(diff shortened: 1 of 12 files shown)


src/k8s_sandbox/_cilium.py
```diff
@@ -0,0 +1,510 @@
...
+    label_selector: str,
+) -> list[dict[str, Any]]:
+    """List one of Cilium's namespaced resources as raw API JSON items.
+
+    See k8s_sandbox._pod.snapshot for why the raw JSON, and why
+    _preload_content needs an ignore.
+    """
+    response = cast(
+        HTTPResponse,
+        custom_objects.list_namespaced_custom_object(  # type: ignore[call-arg]
```
dpolatajko-aisi, line 434: if this is slow or hangs, the deadline won't be honoured - maybe it should have a timeout in terms of remaining deadline budget? there may be other requests where this applies too

PR #260, opened 2026-09-18
Render allowDomains "*" as a world entity, not a wildcard toFQDNs selector

sjawhar:
## The bug

A sandbox whose `allowDomains` names specific domains can **permanently** lose egress to
them while DNS still resolves: `curl` hangs, `POLICY_DENIED` on every SYN, and the
sandbox never recovers for the rest of its life. It happens when another sandbox using
`allowDomains: ["*"]` is scheduled on the same node.

Observed in a staging cluster (Cilium 1.18.4, `identity-allocation-mode=crd`) as three
red CI gates in one day. The failing sandbox made 96 DNS lookups and 108 TCP/443 SYNs
over 2m09s; every SYN was dropped.

## The mechanism

Cilium keeps **one security identity per resolved IP per node**, labelled from the
`toFQDNs` patterns that claimed the name. Each claim re-allocates that identity, and only
the claiming Pod's BPF policy map is rewritten for the new number. A Pod that is still
running keeps the old number, so its egress to that IP starts failing the policy lookup.
Its own retries cannot recover it: the name is already in the node's DNS cache, so the
lookups produce no name→IP *change* to reclaim the identity with.

`matchPattern: "*"` makes this systemic rather than rare, because it claims **every** name
**every** sandbox on the node resolves.

From the node's flow log — one IP for `example.com`, and the identity it was given as
sandboxes came and went (all labelled `fqdn:*`, but a different number each time):

```
15:31:45  dkrscecu  FORWARDED  id=16777220
15:32:22  bkvsga22  FORWARDED  id=16777222   <- fqdn:example.com (scoped sandbox)
15:32:31  ivp6v7j7  FORWARDED  id=16777223
16:15:21  dzdxdmg4  FORWARDED  id=16777225
16:15:37  jlmakjny  FORWARDED  id=16777227
16:15:46  fvybeb5r  FORWARDED  id=16777229   <- re-allocated out from under jlmakjny
16:15:56  jlmakjny  DROPPED    id=16777229   <- jlmakjny's map still allows 16777227
16:16:47  xlxbumh4  FORWARDED  id=16777231
16:16:47  rg3pjdwd  DROPPED    id=16777231   <- scoped sandbox; 108 consecutive drops
```

`jlmakjny` is the mechanism in two lines: it was forwarding fine, a sibling's claim moved
the identity, and it started being dropped. `rg3pjdwd` is the CI failure: a
`allowDomains: ["example.com"]` sandbox that lost the prefix 169 ms *before its first
SYN*, so it never connected at all.

Across 27 sandboxes reaching that host on the same day, 23 installed `fqdn:*` and every
one of them was redundant — their releases also granted the `world` entity.

## The fix

`allowDomains: ["*"]` is a whole-internet grant, and the `world` entity expresses exactly
that without pinning an IP — so it allocates no FQDN identity and cannot disturb anyone.
Render it that way.

| values | before | after |
| --- | --- | --- |
| `allowDomains: ["*"]` | `toFQDNs: ["*"]` | `toEntities: ["world"]` |
| `allowDomains: ["*"]` + `allowEntities: ["world"]` | `toFQDNs: ["*"]` **and** `toEntities: ["world"]`, DNS `*` twice | `toEntities: ["world"]`, DNS `*` once |
| `allowDomains: ["example.com"]` | unchanged | unchanged |
| `allowDomains` + `allowDomainsPorts` | unchanged | unchanged |

The grant is not widened: this repo already treats `allowDomains: ["*"]` as the whole
internet, and the template already skipped SNI/`Host` enforcement for it (`serverNames`
has no allow-all form). Rendering for every list that does not contain `"*"` is
byte-identical — the two scoped rows above were diffed rule-by-rule before and after.

Two sandboxes only contend when their patterns **differ** but match the same name
(`*.example.com` vs `files.example.com`). Identical patterns share one claim and are safe.
That residual is Cilium's per-node identity behaviour rather than something the chart can
fix, so it is written up under Limitations with the two ways to avoid it.

## Verification

- `ruff check` / `ruff format --check` / `mypy` clean; `pytest -m "not req_k8s"` — **439
  passed**, on the rebased tree.
- Rendered policy diffed rule-by-rule for the five value shapes in the table above, before
  and after.
- The post-fix shape is already observable in production: the sandboxes in that cluster
  that grant `world` without a wildcard domain reach the same host via destination
  identity `2` (`reserved:world`), allocate no FQDN identity, and are never subject to
  the eviction above.
- `helm lint` fails on `additional-resources-values.yaml` both with and without this
  change (pre-existing; `helm template`, which the suite uses, is fine).

`test_allow_domains_wildcard_all_skips_identity_enforcement` asserted the old rendering
and is replaced by `test_allow_domains_wildcard_all_becomes_a_world_entity_grant`, which
asserts the new contract plus that DNS still resolves any name. A second test covers the
"named twice" shape.

## Placement: upstream, not fork-local

**Verdict: upstream.** The wrong rendering is in the bundled `agent-env` chart, and the
deployment model it breaks — concurrent releases sharing nodes — is the chart's normal
one. Any consumer hits this, so it belongs here rather than in a fork patch.

**Rejected alternative: fix the caller.** Our runner pairs `allow_entities: ["world"]`
with a redundant `allow_domains: ["*"]`, and deleting that one line would also have
stopped our bleeding. Rejected because it fixes one caller: anyone writing
`x-inspect_k8s_sandbox.allow_domains: ["*"]` in their own compose file reintroduces the
same node-wide selector. The chart owns how a grant becomes a Cilium rule, so the chart is
where a grant that cannot be rendered as an FQDN selector should stop being rendered as
one.

**Class:** upstream-first bug fix in shared chart-rendering logic — no fork-local
divergence introduced, nothing to carry once merged.


sjawhar:
The two failing `k8s` legs are **pre-existing CI infrastructure breakage, not this change.** `main` had not run CI since 2026-09-12, so I ran a control.

**Control:** #261 — upstream `main` `a762e312` plus one inert marker file, nothing else touched. Both `k8s` legs fail there too, with a byte-identical signature:

| signature (occurrences in the helm 3.20.0 k8s leg) | #261 control | #260 this PR |
| --- | --- | --- |
| `containerd-shim-runsc-v1: exec format error` | 4 | 4 |
| `FailedCreatePodSandBox` | 9 | 9 |
| `Helm install timed out` | 4 | 4 |
| `test_can_run_diagnostic_network_issues` | 4 | 4 |

**Cause:** the gVisor shim on the runner is the wrong architecture —

```
FailedCreatePodSandBox: failed to create containerd task: failed to start shim:
start failed: : fork/exec /usr/bin/containerd-shim-runsc-v1: exec format error
```

The diagnostic's sandbox sets `runtimeClassName: gvisor`, so no pod sandbox is ever created, `helm install` hits the 90 s timeout, the eval completes zero samples, and `assert logs[0].results is not None` fails. No CiliumNetworkPolicy is involved: the pod never starts, so policy is never reached. Both legs run with `-x`, so this is the first `req_k8s` test collected and the rest of that suite never executes.

I checked the one way this change could plausibly have caused it: the diagnostic's `helm-values.yaml` does use `allowDomains: ["*"]`, so under this PR its egress comes from the `world` entity instead of a `toFQDNs` rule. That is not the failure — the pod never reaches the point of having policy applied, and the control shows the identical failure with the wildcard `toFQDNs` rule still being rendered.

The non-cluster legs, lint, format, type-check and pre-commit all pass on this PR.

Happy to leave the gVisor/runner arch issue alone here — it wants its own fix and is not in this PR's scope — but flagging it since it currently blocks every `req_k8s` run on `main`.

sylvesterkaczmarek (commented):
Turning allowDomains: ["*"] into a bare world entity changes the port semantics. The normal allowDomains rule is restricted to 80/443, while the toEntities: [world] rule has no toPorts, so this opens every port/protocol to world. Can the world case keep the same 80/443 restriction, with allowDomainsPorts remaining the explicit way to open extras?

sjawhar:
Closing this for now: it was opened ahead of our own review process. The analysis stands and we may re-propose the change after that review; apologies for the noise.

PR #262, opened 2026-09-20
fix: strip helm version output before parsing; declare semver dependency

sjawhar:
## This PR contains

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Other

## Current behaviour

`_validate_helm()` runs `helm version --short` and passes `result.stdout` straight
into `_parse_version()`, which calls `semver.Version.parse()` with no `.strip()`.
The Helm CLI's output ends with a trailing newline (e.g. `"v3.21.3+g1ad6e68\n"`).

python-semver 3.0.4's version regex tolerated that trailing newline (its
end-of-string anchor also matches immediately before a `\n`). python-semver 3.1.0
(released 2026-09-12) tightened that anchor and now raises:

```
ValueError: '3.21.3+g1ad6e68
' is not valid SemVer string
```

Reproduction:

```console
$ uv run --no-project --with semver==3.0.4 python -c "import semver; print(semver.Version.parse('3.21.3+g1ad6e68\n'))"
3.21.3+g1ad6e68
$ uv run --no-project --with semver==3.1.0 python -c "import semver; print(semver.Version.parse('3.21.3+g1ad6e68\n'))"
Traceback (most recent call last):
  ...
ValueError: 3.21.3+g1ad6e68
 is not valid SemVer string
```

Because `validate_prereqs()` runs before any sandbox work, once a dependency
resolver picks up `semver>=3.1.0` (the current PyPI release), every eval using
this sandbox fails immediately at startup, regardless of the installed Helm
version.

Separately, `semver` is imported directly in `_prereqs.py` (`from semver import
Version`) but is **not declared** in `pyproject.toml`'s `dependencies`. It only
resolves today because `inspect-ai` happens to require it (`semver>=3.0.0`,
unbounded) -- an accident of a transitive dependency, not a guarantee.

## New behaviour

- `_parse_version()` strips the input before parsing, so trailing whitespace
  (including the newline Helm always emits) no longer breaks parsing, regardless
  of which semver version is installed.
- `semver>=3.0.0` (matching the floor `inspect-ai` itself uses) is declared as a
  direct dependency in `pyproject.toml`, so it no longer depends on `inspect-ai`
  continuing to require it.

## Other information

Added a regression test (`test_parse_version_strips_trailing_newline`) that feeds
a trailing-newline version string directly through `_parse_version()`, plus a
sibling test without trailing whitespace. Ran, red before the fix / green after
(see below).

### Verification

```console
$ uv sync --extra dev
$ uv run --extra dev pytest test/k8s_sandbox/test_prereqs.py -v          # before fix, semver 3.0.4 (default resolve): 4 passed
$ uv run --extra dev --with semver==3.1.0 pytest test/k8s_sandbox/test_prereqs.py -v   # before fix: 3 failed, 1 passed
$ uv run --extra dev --with semver==3.1.0 pytest test/k8s_sandbox/test_prereqs.py -v   # after fix: 4 passed
$ uv run --extra dev ruff check .        # All checks passed!
$ uv run --extra dev ruff format --check .   # 84 files already formatted
$ uv run --extra dev mypy .              # Success: no issues found in 84 source files
$ uv run --extra dev pytest -m "not req_k8s" -q   # 440 passed, 334 deselected
$ uv run --extra dev pre-commit run --all-files   # all hooks passed
```

Added a `CHANGELOG.md` entry under `## Unreleased`.

No open issue or PR already covers this (searched `semver`, `newline`,
`prereq`/`prerequisite` in this repo's issues/PRs).


sjawhar:
The two `test / pytest (..., k8s)` legs fail with:

```
fork/exec /usr/bin/containerd-shim-runsc-v1: exec format error
event (FailedCreatePodSandBox): ... unable to get OCI runtime for sandbox ...: no runtime for "runsc" is configured
```

That's a gVisor runtime/architecture mismatch on the CI runner, not anything related to
this change (which only touches `_prereqs.py`'s pure-Python version parsing and
`pyproject.toml`'s dependency list). It also fails on unmodified `main` at the same base
commit (`a762e312`) this branch is built on -- a run I triggered on a throwaway
control branch off that exact commit hit the identical failure signature and counts, and
the run that actually pushed that commit to `main` on 2026-09-12 shows both k8s legs
green, so this looks like an intermittent runner-environment issue rather than something
introduced by this PR. The non-cluster legs (lint, format, mypy, pre-commit, and both
non-k8s pytest matrix entries) all pass.

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@
   a failed `ExecResult` rather than raised, so a caller that probes with a user and
   falls back (as inspect-ai does when injecting its sandbox tools) can do so. Naming a
   user that does not exist still raises.
+- Fix startup failing with `... is not valid SemVer string` when a newer `semver`
```
art-dsit, line 29: This can just be 'Fix startup failing with `... is not valid SemVer string`'

art-dsit (approved):

CHANGELOG.md
sjawhar: Done — simplified to just the symptom in 656f2979.

PR #263, opened 2026-09-21
fix(ci): install gVisor binaries from a pinned release

dpolatajko-aisi:
The `gvisor` minikube addon downloads `runsc` and `containerd-shim-runsc-v1` from `releases/release/latest/x86_64/`. gVisor stopped publishing loose binaries there — those URLs now 404 and only `gvisor.tar.{bz2,zstd}` remain. The addon doesn't check the response status, so it writes the 404 XML body to `/usr/bin/runsc` and `/usr/bin/containerd-shim-runsc-v1`, chmods them, and reports success. Every pod using the `gvisor` runtime class then fails with `fork/exec /usr/bin/containerd-shim-runsc-v1: exec format error`, which surfaces in CI as `Helm release did not become ready within 90s`.

This has broken every `k8s` matrix job since 16 September.

Overwrite both binaries from release `20260817.0`, the most recent one that still publishes them individually.

Verified on a local minikube: reproduced the failure (a 205-byte XML `NoSuchKey` body at `/usr/bin/runsc`, pod stuck in `ContainerCreating` with repeated `exec format error`), then applied the pinned binaries — `runsc version release-20260817.0`, the pod went `Running` on the next kubelet retry, and `dmesg` inside it reports `Starting gVisor...`.

Upstream: kubernetes/minikube#23709, fixed by kubernetes/minikube#23715 (milestone 1.40). We can drop this once we're on a minikube release that includes it.


art-dsit (approved):
Might be worth a comment inline that this can be reverted later per the PR description?

dpolatajko-aisi:
> Might be worth a comment inline that this can be reverted later per the PR description?

I will set myself a reminder on the release

PR #264, opened 2026-09-21
fix: accept newline-terminated Helm version output

rasmusfaber:
## Summary

`semver` 3.1.0 was just released and breaks the version parser.

It fails before creating a sandbox because `helm version --short` prints a trailing newline, which the stricter version parser rejects. Strip surrounding whitespace before parsing the version and cover LF/CRLF output with and without the `v` prefix.

## Validation

Reproduced the startup failure using Helm's `v3.21.3+g1ad6e68\n` output with `semver` 3.1.0. The new newline cases failed before the fix; all eight prerequisite tests pass afterward with that version.


rasmusfaber:
This is a dupe of #262. Closing.

